Issue #38, opened 2016-03-08
Component examples


nicolayr:
Examples showing how the various components can or should be used.

nicolayr:
I've already started this for the `TreeView` and `OrganisationUnitTree` components.

Markionium:
For this it would probably also be useful to have stuff like

https://esdoc.org or http://documentation.js.org do some more API like docs.

I experimented a bit with documentation.js for utilizr. Which outputs something like. https://dhis2.github.io/d2-utilizr/

For d2-ui we should probably have the examples properly rendered too with example source. Kind of like 
http://www.material-ui.com/#/components/app-bar perhaps.

PR #48, opened 2016-03-22
#38 sidebar example

caixiaojia:
a simple sidebar example


Markionium:
Thank you @caixiaojia!

Issue #53, opened 2016-03-29
[DataTable] Remove animation from contextmenu


Markionium:
Removing the animation from the context menu does not seem very straightforward.

The `animated` flag on the Popover component does not seem to do the trick even though the docs say it should.

Attempting to disable the animation through style was also not possible as the animation style takes priority over the passed style.

The solution would seem to be to pass a custom version of https://github.com/callemall/material-ui/blob/master/src/Popover/PopoverAnimationDefault.js

PR #57, opened 2016-04-18
Remove animations for context menu for DataTable

adhbh:
Quick (but temporary) fix for https://github.com/dhis2/d2-ui/issues/53

Explanation: 
1. As @Markionium suggested, in material-ui, the animated flag set to false is just effective for exit animation, and not for the the entry animation. The entry animation is still appearing.
2. This is a bug already filed in material-ui. Ref: https://github.com/callemall/material-ui/issues/3140

I looked into the popover code and have come up with this quick fix for DataTable context menu animation. I am working on Material-UI fork to find a better solution but i think for now this is appropriate. What do you think @nicolayr and @Markionium ?


adhbh:
The idea has been taken from this snippet. https://github.com/callemall/material-ui/blob/master/src/Popover/Popover.js#L181

nicolayr:
@adhbh does this actually work as expected? If so it's a clever trick - but also highlights some of the oddities of the Material-UI API.

adhbh:
@nicolayr Yes, this trick removes animation. Here is a quick demo: 

[noanimation.zip](https://github.com/dhis2/d2-ui/files/224869/noanimation.zip)

I agree there are some major issues like this which need to be addressed in material-ui.

nicolayr:
As a side note: We have been discussing the possibility of dropping Material-UI in favor of [Material Design Lite](https://www.getmdl.io/). MDL is a more accurate implementation of the Material design guidelines, and it also doesn't force architecture decisions on us such as using inline styles. On the other hand, it's not made with React in mind, so it's a bit more fiddly to get working correctly.

PR #88, opened 2016-10-03
Add multiple selection to data table

adrianq:
Allows multiple selection in data table. It can be enabled/disabled using the property isMultipleSelectionAllowed, by default it is set to false therefore no changes are expected for current data tables. The only way to execute an action on the selected rows is using the right-click. The three dots at the right side of the row will execute action only for a single row. Regular click works as usual, ctrl+click allows multiple selection.

This PR also fixes some bugs in /FormBuilder.component.test.js due to a typo (Textfield instead of TextField).

We have changed the selection tests in order to make them work with this new property. We have added a couple of tests to test this new feature.

@Markionium Let us know about question or changes @ivanarrizabalaga is the main person behind this feature. He will be more than happy to answer any question you may have.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
     "precommit-hook": "^3.0.0",
     "proxyquire": "^1.7.10",
     "react": "^15.3.0",
+    "react-addons-update": "^15.3.2",    
     "react-addons-test-utils": "^15.3.1",
     "react-color": "^2.2.7",
     "react-dom": "^15.3.0",
@@ -87,8 +88,7 @@
     "draft-js": "^0.7.0",
     "lodash": "^4.12.0",
     "loglevel": "1.4.0",
-    "moment": "^2.11.2",
-    "react-addons-update": "^15.3.2",
+    "moment": "^2.11.2",    
     "react-color": "^2.2.7",
     "recompose": "^0.20.2",
     "rx": "^4.0.8"
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "draft-js": "^0.7.0",
     "lodash": "^4.12.0",
     "loglevel": "1.4.0",
-    "moment": "^2.11.2",    
+    "moment": "^2.11.2",
     "react-color": "^2.2.7",
     "recompose": "^0.20.2",
     "rx": "^4.0.8"
--- a/package.json
+++ b/package.json
@@ -88,6 +88,7 @@
     "lodash": "^4.12.0",
     "loglevel": "1.4.0",
     "moment": "^2.11.2",
+    "react-addons-update": "^15.3.2",
     "react-color": "^2.2.7",
     "recompose": "^0.20.2",
     "rx": "^4.0.8"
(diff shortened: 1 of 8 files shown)


nicolayr:
With regard to UX, do you think that having right click behave differently than the overflow menu (three dots) could be confusing for users?

package.json
```diff
@@ -87,7 +88,7 @@
     "draft-js": "^0.7.0",
     "lodash": "^4.12.0",
     "loglevel": "1.4.0",
-    "moment": "^2.11.2",
+    "moment": "^2.11.2",
```
adrianq: Done! Hope is clearer now!

adrianq:
We had this discussion as well. If the user has many rows selected is 'intuitive' to think that the right click actions are going to be performed to all of them. However the overflow menu is slightly different, right? Because you will go specifically to any of the selected rows and click on the three dots and the action will be applied to all of them. A bit strange... imo the right click is more 'diffuse' (per selection) while the overflow menu is clearly per row. 
On the other hand, it is true what you have just said: to have a different behaviour for the right click than the overflow menu ONLY when multiple lines are selected can be confused for the user.
Difficult to say. We don't have a strong opinion about this and we are not UX experts so up to you... :)

nicolayr:
We discussed this a bit more in depth in the team here, and we have some suggestions:

Since there will be lists both with and without multiple selection, it would be nice to provide some kind of affordance to help users tell the difference. One way of doing that would be to add a column with checkboxes for each item whenever multiple selection is enabled. It would also be nice to have a checkbox in the table header as well, which would make it possible to select/deselect all items and could also act as a sort of status indicator.

When it comes to performing actions on the selected items it might be better to put these in a dedicated area rather than changing the behavior of the existing context menu. One way to do that would be to add an overflow menu (three dots) in the table header which would only be accessible when at least one item in the list is selected.

Does this make sense to you?

adrianq:
@nicolayr Sorry for the late response! Too many things on our plate. 

Your proposal makes all the sense and it would be a much better and clearer implementation. Actually we already discussed something along these lines on our side. However, to be honest, we are very short of time/resources for this project and what we have done so far is enough considering our needs. If, by the end of this project, we finally realise we have some extra time, we will be more than happy to implement these changes ourselves. 

Just note that what we implemented do not change the default behaviour of data tables in DHIS2 and it only adds another behaviour to the component so to say. Anyway, feel free to close this PR, we will open a new one if we have a chance to work on this.

nicolayr:
Thanks for the update @adrianq.

I'm sure we'll eventually run into a situation where we'll need tables with multiple selection as well, so if you don't find the time time to implement these changes we might be able to do it ourselves at some point.

In the meantime, it's probably fine to merge this PR as is. The challenge is that if the final implementation turns out to be incompatible with your current apps, you might find yourself in a situation where you're not able to update. And we would really like to avoid supporting multiple versions of these libraries. The fact that you added tests should help keep that from becoming an issue though.

We do need to test with our current apps first, but assuming that goes well we'll try to get this merged during this release cycle.

varl:
This needs to be adjusted to the new package structure in d2-ui.

PR #89, opened 2016-10-05
Fixed babelrc to give es5 code out and fixed orgUnitTree to be compatible with es5 imports

sultanahamer:
- added es2015 preset
 - removed ModelCollection dependency from orgUnitTree as it causes a cyclic dependency which 
    es5 import is not compatible with. 
 - changed the condition check 
                             from
  if (props.root.children instanceof ModelCollection) 
                                to
  if (!props.root.children instanceof Model) 

we can also change it to something like if(props.root.children.toArray) exists or not.
or any other check like that should be fine.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,5 @@
 {
-    "presets": ["stage-0", "react"],
+    "presets": ["stage-0", "react", "es2015"],
     "plugins": [
         "transform-class-properties",
         "transform-es2015-template-literals",
--- a/src/org-unit-tree/OrgUnitTree.component.js
+++ b/src/org-unit-tree/OrgUnitTree.component.js
@@ -2,7 +2,6 @@ import React from 'react';
 import LinearProgress from 'material-ui/LinearProgress';
 
 import Model from 'd2/lib/model/Model';
-import ModelCollection from 'd2/lib/model/ModelCollection';
 
 import TreeView from '../tree-view/TreeView.component';
 
@@ -42,7 +41,7 @@ class OrgUnitTree extends React.Component {
                 : undefined,
             loading: false,
         };
-        if (props.root.children instanceof ModelCollection) {
+        if (!props.root.children instanceof Model) {
             this.state.children = props.root.children
                 .toArray()
                 // Sort here since the API returns nested children in random order


src/org-unit-tree/OrgUnitTree.component.js
```diff
@@ -42,7 +41,7 @@ class OrgUnitTree extends React.Component {
                 : undefined,
             loading: false,
         };
-        if (props.root.children instanceof ModelCollection) {
+        if (!props.root.children instanceof Model) {
```
Markionium: Did you test if this worked? Since this statement should always return false. As `!props.root.children` will never be an instance of `Model`.

Which basically means that it will never run the `toArray()` when a `ModelCollection` is passed. Perhaps it could be something like:

```js
if (!Array.isArray(props.root.children.toArray) && typeof props.root.children.toArray === 'function') {
```

.babelrc
```diff
@@ -1,5 +1,5 @@
 {
-    "presets": ["stage-0", "react"],
+    "presets": ["stage-0", "react", "es2015"],
```
Markionium: The `es2015` was removed here on purpose to support webpack tree-shaking. (I agree however that we might need to have the other transforms performed).
```json
"presets": ["stage-0", "react", ["es2015", { "modules": false }]],
```

Branch https://github.com/dhis2/d2-ui/tree/v25.0.x still contains es2015, but it would need the cyclic dependency fix you mentioned.

src/org-unit-tree/OrgUnitTree.component.js
sultanahamer: Yup, absolutely. I have run the code once and ran tests but missed this some how. My intention was to check if this is not of type Model. But even that could go wrong when we get a d2 Model object whose children has some object not being a model collection. My initial check was checking if toArray exists on the children but I felt it not being a good fix. Shall I go with the above check then with one more condition on top to check if children exists in the first place. 
if (props.root.children && !Array.isArray(props.root.children.toArray) && typeof props.root.children.toArray === 'function')

.babelrc
sultanahamer: I have checked this on the v25.0.x branch and the lib being generated is looking fine.

src/org-unit-tree/OrgUnitTree.component.js
sultanahamer: will create a pull request for 25.0.x branch.

PR #110, opened 2017-03-15
Feature/sharing dialog

kjesvale:


src/sharing/Sharing.component.js
```diff
@@ -1,120 +1,199 @@
-import { PropTypes, createClass, default as React } from 'react';
+/* eslint react/jsx-no-bind: 0 */
```
Markionium: I understand why you did it but lets try to not override the config. There are a few other places where you disable some rules which could have been solve by using different constructs.

src/sharing/SharingDialog.component.js
```diff
@@ -1,44 +1,251 @@
...
+    const userGroupAccesses = !apiObject.userGroupAccesses ? [] : apiObject.userGroupAccesses.map(
+        access => transformAccessObject(access, 'userGroup'));
+
+    const userAccesses = !apiObject.userAccesses ? [] : apiObject.userAccesses.map(
+        access => transformAccessObject(access, 'user'));
+
+    const combinedAccesses = userGroupAccesses.concat(userAccesses);
+
+    return {
+        ...apiObject.user && {
```
Markionium: This construct is somewhat confusing. 

It will leave out the `authorOfSharableItem` when `apiObject.user` is `undefined` or `null` is that the intended result or is it primarily to guard against `apiObject.user` to be falsy?

src/sharing/SharingDialog.component.js
```diff
@@ -1,44 +1,251 @@
...
+        return apiInstance.get('sharing/search', { key: searchText })
+            .then((searchResult) => {
+                let transformedResult = searchResult.users.map(
+                    user => transformAccessObject(user, 'user'));
+
+                transformedResult = transformedResult.concat(
+                    searchResult.userGroups.map(
+                        userGroup => transformAccessObject(userGroup, 'userGroup')));
+
+                return transformedResult;
```
Markionium: You could return the result from the concat directly.

src/sharing/SharingDialog.component.js
```diff
@@ -1,44 +1,251 @@
...
+              return message;
+          })
+          .catch(({ message }) => {
+              console.warn('Failed to post changes.');
+              console.warn('SERVER SAID:', message);
+          });
+    }
+
+    loadObjectFromApi() {
+        getInstance().then((d2) => {
```
Markionium: We could grab this one from the context. I think i might have suggested you to do it like this but sicne we already put `d2` on the context here we might aswel grab it from there.

src/sharing/SharingDialog.component.js
```diff
@@ -1,44 +1,251 @@
...
+                });
+        });
+    }
+
+    restoreObjectStructure(transformedObject) {
+        const userAccesses = [];
+        const userGroupAccesses = [];
+
+        // eslint-disable-next-line array-callback-return
+        transformedObject.accesses.map((access) => {
```
Markionium: If you're not going to use the result of the `.map` you can use `.forEach` (which would also get rid of the eslint error ;))

src/sharing/SharingDialog.component.js
```diff
@@ -1,44 +1,251 @@
...
+                ? userAccesses.push(apiAccess)
+                : userGroupAccesses.push(apiAccess);
+        });
+
+        return {
+            meta: this.state.apiObject.meta,
+            object: {
+                ...this.state.apiObject.object,
+                ...userAccesses && { userAccesses },
+                ...userGroupAccesses && { userGroupAccesses },
```
Markionium: This could be written as just `userAccesses` and `userGroupAccesses`? The variable will always be initialized as you do that on lines 148 and 149. 

```js
object: {
    ...this.state.apiObject.object,
    userAccesses,
    userGroupAccesses,
```

src/sharing/SharingDialog.component.js
```diff
@@ -1,44 +1,251 @@
...
+                onClick={this.closeSharingDialog.bind(this)}
+            />,
         ];
 
-        return (
+        const loadingMaskStyle = {
+            position: 'relative',
+        };
+
+        return this.state.objectToShare ? (
```
Markionium: I know this is a remnant of my old stuff and that we have this in multiple places but perhaps we should rewrite it to be like the following to make it somewhat cleaner?

```js
if (this.state && !this.state.objectToShare) {
    return (<LoadingMask style={loadingMaskStyle} size={1} />); 
}

return <Dialog ... />
```

test/sharing/AccessMaskSwitches.component.test.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+/*import React from 'react';
```
Markionium: Probably an oversight, but we should probably remove this commented code.

test/sharing/UserGroupAccesses.component.test.js
```diff
@@ -58,3 +101,5 @@ describe('Sharing: UserGroupAccesses component', () => {
         expect(() => accessMaskSwitches.simulate('change', 'rw------')).not.to.throw();
     });
 });
+
+*/
```
Markionium: Can be removed :)

webpack.config.js
```diff
@@ -99,12 +100,11 @@ module.exports = {
                 'NODE_ENV': JSON.stringify(isProduction() ? 'production' : 'development'),
             },
         }),
-        isProduction() ?
+        isProduction() ? null :
```
Markionium: Thanks ! :P

PR #114, opened 2017-09-11
Feature/button

turban:
New Button component wrapping Material UI Button v0.19, but mirroring the Material UI v1 API.

### DIFF
--- a/src/button/Button.js
+++ b/src/button/Button.js
@@ -49,7 +49,7 @@ Button.propTypes = {
     fab: PropTypes.bool,
     disabled: PropTypes.bool,
     color: PropTypes.string,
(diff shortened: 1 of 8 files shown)


src/button/Button.js
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     *  If true, the button will be disabled
+     */
+    disabled: PropTypes.bool,
+
+    /**
+     * The theme color of the button ('default', 'primary', 'accent')
+     */
+    color: PropTypes.string,
```
nicolayr: Here you could also use the `oneOf` prop type to check specifically for the valid values:

`PropTypes.oneOf(['default', 'primary', 'accent'])`

PR #116, opened 2017-09-14
Fix eslint errors on src directory

jenniferarnesen:
Upgrade to latest eslint-config-dhis2 and run eslint src --fix


jenniferarnesen:
Upgrade to latest eslint-config-dhis2 and run eslint src --fix

PR #118, opened 2017-09-15
Fix eslint no-unused-vars errors

jenniferarnesen:


jenniferarnesen:
We should discuss if all of these are ok to merge.

PR #124, opened 2017-10-23
Transpile import statements to commonjs for the build

jenniferarnesen:
Apps that have d2-ui as a dependency run into problems when running tests, since packages in node_modules are not transpiled, returning an error that the "import MyComponent from 'd2-ui' " is a syntax error. It is possible to make workarounds for this in the apps (maintenance-app does this, see jest.config.js) but it is not possible to do a workaround for any apps created with create-react-app.

Therefore, it has been decided that d2-ui will transpile import statements to commonjs syntax, as part of the build/npm package.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -7,6 +7,7 @@
         "transform-es2015-function-name",
         "transform-es2015-arrow-functions",
         "transform-es2015-block-scoped-functions",
+        "transform-es2015-modules-commonjs",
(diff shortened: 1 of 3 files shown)


Birkbjo:
This can be fixed by using latest version of node, no?

jenniferarnesen:
Lets not do this now. We can revisit another time.

PR #128, opened 2017-10-30
Feature/select field

turban:
Includes a new wrapper around MUI SelectField, including examples and tests. 

There are also smaller fixes in Button and TexField wrappers.


turban:
Only reason is that the event handlers are set as required, and this avoids a warning in the test console.

PR #131, opened 2017-11-01
Feature/Tabs

turban:
This pull request includes wrappers for Material UI Tabs and Tab. Tabs is wrapped in a functional component, while Tab is wrapped in a class component to maintain state (to allow tab switching). 

The pull request also includes a utility function to create class names in the wrappers.


turban:
Thanks for your feedback @Stevenah! Now fixed.

PR #135, opened 2017-11-14
Add an SvgIcon component

jenniferarnesen:
This component support a limited set of icons that are expected to be used in the near future. More can be added later.  The component currently does not support setting the color or style of the icon in any way.

The example page displays all the supported icons.


nicolayr:
Would it be useful to show all the explicitly supported icons in the example app?

That way we would also have an easy way to look up which icons are already being used in other apps, which should make it easier to ensure that we use the same icons consistently across apps.

turban:
Looks fine, a few comments:
- Include documentation of props we want to expose
- No big deal, but maybe use PascalCase rather than camelCase for icon names, to be in line with Material UI v1: https://material-ui-next.com/style/icons/#svg-material-icons
- Congrats with 100% test coverage 👍

jenniferarnesen:
@turban 
I've updated the PR as you requested. I'll go ahead and merge.

PR #137, opened 2017-11-23
Feature/chip: Wrapper for the Chip component

janhenrikoverland:


turban:
Feedback: 
- Add unit tests ("yarn coverage" until you reach 100%)
- Use ternary operator or && to disable events? (reads better)
- Use the SVG icons that @jenniferarnesen is working on: https://github.com/dhis2/d2-ui/tree/feature/add-svg-icons
- Missing semicolon on line 45 in Chip.js
- We should find a way to define a shared theme/colors
- Use "d2-ui-chip-selector" in the prop types docs

janhenrikoverland:
2) Ternary would return either the function or 'false' which violates the MUI prop type.

turban:
Looks fine now and congrats with 100% test coverage! 🎉

Maybe just replace:

style: style,
labelStyle: labelStyle,

with 

style, 
labelStyle

in src/chip/Chip.js

PR #139, opened 2017-11-27
Feature/Passed reference of formModel to validation. Refactored react.proptypes

paragateoslo:


src/form-fields/DatePicker.component.js
```diff
@@ -61,6 +68,17 @@ DatePicker.propTypes = {
     onChange: PropTypes.func.isRequired,
     dateFormat: PropTypes.string.isRequired,
     allowFuture: PropTypes.bool.isRequired,
```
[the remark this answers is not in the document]
nicolayr: Also the `dateFormat` prop only has 2 valid values: `'dd-MM-yyyy'` and `'yyyy-MM-dd'`. It would be better to use `PropTypes.oneOf([ ... ])` validation.

src/form-fields/DatePicker.component.js
```diff
@@ -6,14 +6,20 @@ import { DatePicker as MuiDatePicker } from 'material-ui';
...
-        this.maxDate = props.allowFuture ? undefined : new Date();
-        this.props = props;
-        this._onDateSelect = this._onDateSelect.bind(this);
-        this._formatDate = this._formatDate.bind(this);
+        const {
+            allowFuture,
+            dateFormat,
+            ...other
+        } = props;
+        this.other = other;
```
nicolayr: This props magic should probably go in `render()` instead of in the constructor :)

package.json
```diff
@@ -8,7 +8,7 @@
     "examples": "npm run start",
     "coverage": "jest --coverage",
     "start": "webpack-dev-server",
-    "test": "jest",
+    "test": "jest --env node",
```
nicolayr: Is this necessary? Can it cause any problems?

src/form-fields/DropDown.component.js
```diff
@@ -59,9 +61,20 @@ DropDown.propTypes = {
...
+        PropTypes.number,
+        PropTypes.bool,
+    ]),
+    menuItems: PropTypes.oneOfType([
+        PropTypes.array,
+        PropTypes.object,
+    ]),
+    onFocus: PropTypes.func,
+    onBlur: PropTypes.func,
     includeEmpty: false,
```
nicolayr: If you add a default value for `noOptionsLabel` here, the code in `render()` can be simplified as well since it will be safe to assume that it will always have a value.

src/forms/AsyncValidatorRunner.js
```diff
@@ -8,9 +8,9 @@ import { Observable, Subject } from 'rxjs';
  * or with `{ isValid: false, message: '<error_message>'`} when one of the validators failed. The `message` property
  * contains the value that the failed validator function rejected with.
  */
-function runValidatorFunctions(validatorFunctions, value) {
+function runValidatorFunctions(validatorFunctions, value, formModel) {
```
nicolayr: Strictly speaking the `formModel` parameter is not a model but an object representing the state of the form. If you instead call it `formState` for example, it'll be easier to understand what's actually going on. This applies to the rest of this file as well of course.

jenniferarnesen:
@paragateoslo You'll probably want to merge in master and fix conflicts now. The changes that Nico was referring to earlier are now merged to master.

PR #140, opened 2017-11-28
Feature/react16: Eliminate deprecated React code

janhenrikoverland:


janhenrikoverland:
Merging this PR basically will:

- Migrate from React.createClass to React.Component extended ES6 classes
- Use the prop-types package instead of React.PropTypes
- Declare propTypes, contextTypes, defaultProps etc outside of the component
- Declare initial state as an object (state = {}) inside the class
- Get translations via React context rather than importing it as a mixin
- Fix some syntax issues like no newline between class functions

jenniferarnesen:
Additional changes:
- remove warnings from running tests
- remove deprecated files
- upgrade major version to 29

jenniferarnesen:
I've separated this PR into 2 new pull requests, 1 for the preparations for react 16, and another for the removing of deprecated files.

PR #143, opened 2017-12-04
Remove deprecated index.js files

jenniferarnesen:
Remove unused `Auth.mixin.js` as well as `index.js` files that have been deprecated for several releases.


jenniferarnesen:
@nicolayr Could you just note in the description why the Auth.mixin is removed? And also just confirm that we don't need to do a major upgrade for these changes?

nicolayr:
Just for the record: The fact that we're removing `index.js` files now does not necessarily mean that we should never have `index.js` files in the future.

The problem with the deprecated approach was that some folders contain multiple components so that when importing `index.js` it's not clear exactly which component is being imported. A more consistent folder structure where each folder contains only one public component could be a better solution to the problem.

PR #156, opened 2017-12-20
Add unit tests for AppWithD2 component

jenniferarnesen:


Markionium:
Hi :)

Good you added some tests. 

There is another similar component in the same folder called D2UIApp,  which is somewhat more advanced. You might prefer using that instead for apps. 

It also seems there is a copy paste error in the AppWithD2 file that was already there. It's component name is DataTableHeader, but it seems safe to change. :)

jenniferarnesen:
Hey @Markionium 👻 
Thanks for the tip. I noticed this component is used in Maintenance app. I'll add a point to the d2-ui discussion on starting to deprecate this component in favor of D2UIApp.

Yeah, I noticed the "DataTableHeader" too.

Thanks for checking in :)

PR #162, opened 2017-12-21
Task/prefer svgicon in sharing

edoardo:
I'm using the SharingDialog component in the new FavoritesDialog component (WIP) and got broken UI due to missing material icons font in my test app.
With nico we decided to use SvgIcon instead.

Replace FontIcon with SvgIcon in the Sharing component.
Added the missing SVG imports in the SvgIcon component.

Sorry for the huge diff, do we have special settings for prettier? I thought we run pretty much with the defaults, but I might have to tweak the settings in my editor.

### DIFF
--- a/examples/svg-icon/index.js
+++ b/examples/svg-icon/index.js
@@ -32,12 +32,108 @@ const icons = (
                 <SvgIcon />
             </div>
             <div style={cardStyle}>
+                <h3>Icon with custom defined path</h3>
+                <SvgIcon>{children}</SvgIcon>
+            </div>
+            <div style={cardStyle}>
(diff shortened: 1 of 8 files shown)


jenniferarnesen:
@edoardo Can you add the new icons to the svgicon example?

jenniferarnesen:
@edoardo Regarding prettier, it is not set up in this repo. If you are running prettier in your workspace, it looks like you might need to add a few configurations, like Arrow Function Parentheses, and line width. Here is a link to the options: https://prettier.io/docs/en/options.html

edoardo:
@jenniferarnesen Examples and tests added.

PR #167, opened 2018-01-11
Add new TextField properties to support multiline fields

jenniferarnesen:
Also adds properties to support hint text and fullwidth. Note that the props passed in to the d2-ui component comply with the MUI 1.0 property naming. (multiline, placeholder)

I rearranged the properties alphabetically which makes the code changes appear bigger than they actually are. Apologies for that.


jenniferarnesen:
I'm a little mixed on providing default properties. The linter wants it, but the MUITextField has default properties already, so its not really necessary.

PR #174, opened 2018-01-24
Add information about contributing to the README

jenniferarnesen:


README.md
```diff
@@ -9,19 +9,72 @@ UI Component library for developing [DHIS2](http://www.dhis2.org) applications.
...
+
+Some of the examples require a local DHIS2 instance. Most developers have their DHIS2 instance running on http://localhost:8080. If this is the case for you, then the examples should just work.
+
+
+### Publishing to npm
+
+The d2-ui library is published as an npm package: https://www.npmjs.com/package/d2-ui
+
+#### A note about d2-ui semver
+The d2-ui version tracks the DHIS2 version. So, when DHIS2 version 2.28 is released, then d2-ui will be upgraded to v28.0.0. Therefore, all version upgrades prior to the next DHIS2 release are generally patch upgrades, even if the changes are considered breaking.
```
varl: I see that we cannot adjust the major version of d2-ui when releasing breaking changes, but we should push minor version bumps for "bigger" stuff which are potentially breaking, and just use patch releases for non-breaking stuff.

We should strive to ensure that d2-ui doesn't accidentally and automatically break for consumers of the lib.

README.md
```diff
@@ -9,19 +9,72 @@ UI Component library for developing [DHIS2](http://www.dhis2.org) applications.
...
+
+#### Steps to publish
+
+After your pull request with functional changes has been merged to master, switch to the master branch locally and pull to get all the latest changes. The version upgrade will be done directly on the master branch:
+
 ```
-npm install --save d2-ui@24
+git co master
```
varl: I don't think `co` is a default binding in Git, we should be more specific in the docs and use `checkout`.

README.md
jenniferarnesen: I'm actually not completely sure what conventions the previous team members used (@Markionium @nicolayr ).  But I guess we can define the conventions according to what we think makes the most sense? Using minor upgrades to indicate a breaking change could definitely be one of those things.

README.md
jenniferarnesen: Agree. Will fix..

README.md
jenniferarnesen: The other thing I am not sure of is when exactly we do the major version, and how all the apps with d2-ui as dependency get the correct version.

varl: I think each app needs to update itself with the new version, if they need it.

varl:
@jenniferarnesen LGTM, merge when you are ready. 👍

PR #187, opened 2018-02-15
Filter for OrgUnitTree

joaoantunes87:
Suggestion to filter the Organisation Units to show. There are some situations where we do not want to show all children (OrgUnits), for example when we have an Organisation Units list available for a specific Data Set.

This approach receives a prop of organisation units to include when rendering children. I am not saying this is the best approach, however given what we have right now I think it does not have performance problems, neither it changes much code.

@varl could you review it.

### DIFF
--- a/examples/org-unit-tree/index.js
+++ b/examples/org-unit-tree/index.js
@@ -67,6 +67,15 @@ function OrgUnitTreeExample(props) {
                 </Card>
                 <Card style={styles.card}>
                     <CardText style={styles.cardText}>
+                        <h3 style={styles.cardHeader}>Plain OrgUnitTree with filter</h3>
+                        <OrgUnitTree
+                            root={props.root}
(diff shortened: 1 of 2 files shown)


HendrikThePendric:
Hi @joaoantunes87. This approach has some limitations. To reproduce these, please do the following:
- There is an OrgUnit called Dibia with id `ZiOVcrSjSYe`. It is under Port Loco, so three levels deep.
- Its parent, Port Loco, is not in the list of orgUnitsIdsToInclude in the example
- If you add the Dibia ID to the list of orgUnitsIdsToInclude in the example, it will NOT display because the parent is already filtered out.

You could argue that this is the responsibility of the filter function. And that the filter needs to produce a full list of IDs that always includes the needed parents.

However, I would argue that we should switch to a property called pathsToInclude that takes care of the filtering. Not sure if I should do this in your fork or here though....

joaoantunes87:
Ah right. I had not thought about that case. I was using path in my first approaches, however I decided to use ids because it seemed to be the right choice. However, I do not see any problem of using paths.

About the changes I think you could use my fork. Maybe @varl is the best person to answer it.

joaoantunes87:
Solved [here](https://github.com/dhis2/d2-ui/pull/188).

PR #188, opened 2018-02-15
Filter by path instead of ID

HendrikThePendric:
Alternative suggestion for filtering the tree using path instead of id. By using this approach the parents of deeply nested children will always be displayed in the tree.

Another additional example, is that it now uses the same structure as the initiallyExpanded property. Which means you can easily expand the filtered tree too.

### DIFF
--- a/examples/org-unit-tree/index.js
+++ b/examples/org-unit-tree/index.js
@@ -67,6 +67,15 @@ function OrgUnitTreeExample(props) {
                 </Card>
                 <Card style={styles.card}>
                     <CardText style={styles.cardText}>
+                        <h3 style={styles.cardHeader}>Plain OrgUnitTree with filter</h3>
+                        <OrgUnitTree
+                            root={props.root}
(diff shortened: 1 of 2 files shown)


joaoantunes87:
Hi @HendrikThePendric, 
I think it is better this way. I have tested it and everything seems ok as well the code. Agree with using path instead of id, once it is already being used for initiallyExpanded property.

In my opinion we only have to gain on merging this PR. However, I think in the future we will need to check if it will not be a problem for big Org Unit Trees with a lot of nodes like 80 thousands as someone was mentioning on slack as a possibility in some cases.

joaoantunes87:
Just adding [here](https://github.com/dhis2/d2-ui/pull/187) the first [PR](https://github.com/dhis2/d2-ui/pull/187). In case another person not involved in the initial conversation needs to see how the conversation started.

varl:
I think this looks good and the option to filter orgunit trees is a good feature. 

We know that there are at least two systems in the wild which have 80k+ nodes. I don't see any reason that this approach would be problematic but I am curious about if you guys have any ideas on how to stress test this?

joaoantunes87:
Right now, I do not have any ideas on how to do the stress tests. If any idea come to mind I will keep you updated.

Thanks @HendrikThePendric and @varl for your help. 
Let me know when we are able to use d2-ui with this update.

Issue #189, opened 2018-02-15
Period Picker: second input does not update on UI


joaoantunes87:
On PeriodPicker Component when we have two inputs, for example year and week, the value of the second one does not get updated on the UI, but it triggers the **onPickPeriod** callback well.

To reproduce it I use the Period Picker example at this project [here](https://github.com/dhis2/d2-ui/blob/master/examples/period-picker/index.js).

The problem can also be seen on this image:
![image](https://user-images.githubusercontent.com/1324999/36256282-28773a4e-124a-11e8-9746-77b52f16c0e7.png)

PR #190, opened 2018-02-15
d2-ui issue 189 bug fix for PeriodPicker

joaoantunes87:
Bug fix in regard [this](https://github.com/dhis2/d2-ui/issues/189) issue:
I have noticed the problem. However I do not have enough knowledge about d2-ui code to understand why the **reset** (see diff code) I removed was there. 

Better anyone with more knowledge on d2-ui to check it.

@varl could you give a check on it, please?


jason-p-pickering:
@varl can you review or assign to someone. This is needed for a bug we found as part of the data admin app.

varl:
Hm, I don't know why we clear the state that way in the callback handler, but either way it doesn't seem like the best way to do it.

PR #192, opened 2018-02-23
Feature/favorites and sharing

varl:


packages/sharing/src/UserSearch.component.js
```diff
@@ -73,6 +74,9 @@ class UserSearch extends Component {
             this.props.addUserAccess({ ...selection, access: accessObjectToString(this.state.defaultAccess) });
         } else this.props.addUserGroupAccess({ ...selection, access: accessObjectToString(this.state.defaultAccess) });
```
edoardo: This looks like a leftover that needs to be removed...

varl:
The tests run locally now (let's see what Travis says).

Gonna fix the examples too.

varl:
Found the issue with Travis. The way I fixed it doesn't scale so need to add a better script for managing deps (e.g. installing) for standalone components.

PR #196, opened 2018-02-27
changed the PTEA expression format to actually add an id an not name

john-melin:
issue: https://jira.dhis2.org/browse/DHIS2-2625

- Changed the programOperandSelector to handle an array instead of a collection as it has always been. 
- Linting of GroupEditor and GroupEditorWithOrdering that was causing warnings.

### DIFF
--- a/src/expression-manager/ProgramOperandSelector.js
+++ b/src/expression-manager/ProgramOperandSelector.js
@@ -51,7 +51,7 @@ class ProgramOperandSelector extends Component {
     };
 
     componentDidMount() {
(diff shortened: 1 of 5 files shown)


src/expression-manager/ProgramOperandSelector.js
```diff
@@ -108,8 +106,7 @@ class ProgramOperandSelector extends Component {
     }
 
     onProgramTrackedEntityAttributeSelected = (value) => {
-        const programTrackedEntityAttributeFormula = ['A{', value, '}'].join('');
-
+        const programTrackedEntityAttributeFormula = ['A{', this.state.selectedProgram, '.', value, '}'].join('');
```
kjesvale: Change to string template: `A{${this.state.selectedProgram}.${value}}`

PR #202, opened 2018-03-11
Refactor build process

Birkbjo:
This is WIP.

- Sharing dialog is reduced in size for umd build: 650kb to 240kb (mostly due to RxJS stuff).
- Sharing dialog is distributed in es, cjs and umd.
- Tests can be run inside Sharing directory. This is probably not necessary, and I think we should just have one "test-setup" in the root.
- es distribution will be used by default as we write our apps with import-statements. This also makes it possible for the apps to do treeshaking.

Still need to do this for the entirity of d2-ui (and d2) as well.

- [x] Everything should be built to es, cjs and umd
- [x] Find a solution to install dependencies for all packages (lerna, yarn workspaces)
- [x] Should probably move "d2/src" to "packages/d2-ui" and rename to d2-ui-core. (needed for lerna and workspaces). 
- [x] Compose a list of components that should be moved out of the core. (E.g. Example app).
- [x] Replace the haunting compass with node-sass

### Some notes: 
"main" and "module" can be used in package.json to define the main-entry point of the package, where each will be used by commonjs imports (require()) and es-imports (import ...), respectively. As we copy the package.json and change it in the move-script, we can rename these fields. My first thought was that this would make yarn link-ing while working on a component in an app much easier. As we could point the entry point to "src/index.js" and then just link with the app being worked on, and thus skip the build-step in the component before seeing the changes in the app. However, this file will not be transpiled by babel (and no easy "plug-and-play" solution seem to exist, see https://github.com/webpack/webpack/issues/1643 and https://github.com/babel/babel-loader/issues/149). 
Therefore I see no reason to keep the "copy package.json to build-folder and publish from build"-procedure necessary. The desired behaviour of only publishing some files should be able to be done through a combination of package.json "files", "main" and "module"-fields. 

We should probably look more into lerna and learn(a) its implications (version changes on each package etc.). It really seem to simplify a lot of the problems with the current approach that is a bit hacky (like manually installing dependencies in packages before tests). Lerna and yarn workspaces would do this for us. It would also make changes using "d2-ui"-example app much easier, as you would not need to worry about transpilation of extracted-components.

For good measure:
The workflow for editing the component while linking with an "outside"-app is like this (and will probably stay like this):

- cd to e.g "sharing" and "yarn link"
- run "yarn watch"
- cd to app and yarn link "d2-ui-sharing"
- *Hack in component*
- Changes should be picked up in app due to babel-watching the "sharing"-files for changes


packages/sharing/webpack.config.js
```diff
@@ -1,25 +1,40 @@
...
+const webpack = require('webpack');
+const path = require('path');
+const rxPaths = require('rxjs/_esm5/path-mapping');
 
 module.exports = {
     entry: './src/index.js',
     output: {
         path: path.resolve(__dirname, 'build'),
         filename: 'index.js',
         library: 'SharingDialog',
```
varl: We should do the same for Favorites as well.

varl:
Components to move out of `d2-ui-core`:

- HeaderBar (@varl is working on this)
- OrgUnit* components (depends on d2)

Other improvement suggestions to consider:

- Use styled-components instead of SASS to bundle the style with the componets, implement themes, etc.

varl:
HeaderBar and Favorites now have the same _modus operandi_ as Sharing.

varl:
I tried implementing Lerna, but they [endorse using the standard NPM mechanisms](https://github.com/lerna/lerna/issues/901) for publishing a package which gives us ugly import paths: `import foo from 'bar/dist/Foo'` and not the nicer `import foo from 'bar/Foo'` which e.g. Material UI has. Not a perfect fit for us IMO.

So for now I at least added Yarn Workspaces which take care of all the build/test dependencies in the root package, so the packages only define the dependencies they need runtime. That means one `.babelrc` in the root workspace, and all e.g. Babel, Jest, Webpack versions are handled in the root `package.json`. Each package brings its own `webpack.config.js` to the party though, so they can be customised per sub-package (headerbar, sharing, favorites, etc.).

As of now the workflow is as follows:

1. `yarn install` is run in the root package (d2-ui)
2. `yarn build` and `yarn link` is run for each package.

Problems to solve:

Versioning packages. The packages can be bumped in version independently, but the root `package.json` needs to be updated to the latest of them. The packages are published independently to `npm` by using:

`cd packages/headbar && npm publish build` where `build` is the directory which contains the built package.

varl:
Using workspaces the `node_modules` is now 230 MB instead of 315 MB and takes a lot less time to link. I think we can shave more here as well when we get further along.

packages/sharing/webpack.config.js
varl: Did it.

varl:
>Therefore I see no reason to keep the "copy package.json to build-folder and publish from build"-procedure necessary. The desired behaviour of only publishing some files should be able to be done through a combination of package.json "files", "main" and "module"-fields.

As we have discussed publishing the `build` dir is to get a flat package structure with easy and predictable import paths. In NPM we can use `npm publish <dir>` so we can publish the package in question from the root dir (skips the `cd <dir>` step, which I agree is messy).

varl:
# D2 reliant components

- [ ] HeaderBar 
- [ ] DataTable
- [ ] OrgUnitTree component
- [ ] OrgUnit Selection
- [ ] Formula Editor
- [ ] Legend
- [ ] GroupEditor
- [ ] Sharing
- [ ] Control Bar
- [ ] FavoritesDialog

varl:
Now that I am sitting here with 11 packages under `packages/` I have evaluated writing our own little tool to do some simple tasks for all of the packages like packaging, releasing, testing, and building, and spent some time reading Lerna's source code to compare and see how much work it would be to do something similar but smaller in scope.

The trade-off between keeping our import style the way we want vs. the pain of managing versions, dependencies, and packages leads me to weigh in favour of Lerna.

Two important notes:

## Re-export file

We need to re-export everything in d2-ui which we want to expose to the consumer of the library. This is not necessarily a bad thing as it allows us to change the structure and organisation within d2-ui without affecting the interface for the consumer.

## Named imports from the re-export file

We will have to change the way we import from e.g. d2-ui to named `import`s:
```
//  old
import Button from 'd2-ui/lib/Button'

// new
import { Button } from 'd2-ui'
```
The new style still supports tree-shaking if we use named imports.

varl:
So I don't lose these two links:

- https://github.com/semantic-release/semantic-release
- https://github.com/atlassian/lerna-semantic-release

Used to automate the build and release process, along with creating automatic changelogs and tracking breaking changes.

varl:
Only two components (the heaviest ofc) remaining to put into the examples application, but postponing that for now:

- [ ] OrgUnit Select
- [ ] OrgUnit Tree

varl:
Only two components (the heaviest ofc) remaining to put into the examples application, but postponing that for now:

- [ ] OrgUnit Select
- [ ] OrgUnit Tree

varl:
With UMD build:

`Done in 299.53s.`

Without UMD build:

`Done in 20.15s.`

I think we should only run UMD builds on the deploy command so we don't have to wait 5 minutes (locally) and 15 minutes (on travis) to run a build for each PR.

PR #210, opened 2018-04-03
feature: options component for configuring data visualizer

joakimia:
Pull Request for adding re-usable options component for configuring data visualizer's graphical output.
The feature is using material-ui-next (v1) components.
Top-Level component is ChartOptions.
As of now there's minimal styling added.


edoardo:
I don't know if this has been discussed already, but I'm not sure these components belong to d2-ui.
They are very specific to the new DV app, so I would rather have them there (https://github.com/dhis2/data-visualizer-app).

varl:
I think that the conclusion here is to move it to the data visualizer app for now. Closing.

PR #211, opened 2018-04-06
Add FavoritesMenu component

edoardo:
The translation dialog is not working properly, the language drop-down does not show in the proper place and it seems it needs 2 clicks in order to load the languages.

@varl I wanted to push the branch so you can have a look and give feedback on the whole component.

The translations for other components are also missing.
Not sure how to solve that, I guess we need to work on changing the way the translations are handled for these components using the new i18n library.


packages/favorites-menu/package.json
```diff
@@ -0,0 +1,33 @@
...
+            "cross-env BABEL_ENV=commonjs babel src --out-dir build/cjs --ignore spec.js",
+        "build:es": "cross-env BABEL_ENV=es babel src --out-dir build/es --ignore spec.js",
+        "build:umd": "cross-env && webpack -p",
+        "watch": "npm run build:es -- --watch",
+        "clean": "rimraf build"
+    },
+    "dependencies": {
+        "d2-ui": "^29.0.18",
+        "d2-ui-favorites": "^1.0.4",
+        "d2-ui-sharing": "^1.0.1",
```
varl: these should be `d2-ui-favorites-dialog` and `d2-ui-sharing-dialog`

edoardo:
Added basic tests for the menu item components.

PR #212, opened 2018-04-06
Added isInteger validator, fixed GroupEditorWithOrdering warning wher…

john-melin:
- Added isInteger validator
- Fixed GroupEditorWithOrdering warning where assignedItemStore could only be PropType object.


jenniferarnesen:
@john-melin master branch has changed big time, so you'll have to apply these changes to the new d2-ui structure.

PR #216, opened 2018-05-04
Move d2-utilizr dep from root to the packages that need it

edoardo:
Fixes TECH-43.

- add d2-utilizr dependency in each package that uses it
- remove d2-utilizr from d2-ui root dependencies


varl:
Published to npm.

PR #217, opened 2018-05-04
Change font weight and text on permission picker

kjesvale:
Changes requested by Lars.

- Change font weight in permission picker from 400 to 500
- Change text in data sharing field


varl:
LGTM.

PR #220, opened 2018-05-08
Rename FavoritesMenu component to FileMenu

edoardo:
Fixes TECH-51 (and perhaps TECH-30)

- Favorites should not be used anymore as it is a confusing term. Replace the FavoritesMenu component with the FileMenu
- Added i18n


packages/file-menu/package.json
```diff
@@ -0,0 +1,44 @@
...
+        "build"
+    ],
+    "license": "BSD-3-Clause",
+    "author": "Edoardo Sabadelli <edoardo@dhis2.org>",
+    "peerDependencies": {
+        "d2": "^29.1.1",
+        "react": "^15.0.0 || ^16.0.0",
+        "react-dom": "^15.0.0 || ^16.0.0"
+    },
+    "scripts": {
```
varl: Don't we need the string extraction scripts and deps here?

https://github.com/dhis2/d2-i18n#in-apps-packagejson

edoardo: Added deps and scripts and generated the pot file.

varl:
If you fix the conflicts we can merge this and I can do the release process.

PR #222, opened 2018-05-14
refactor search field reference

adeelshahid:
- fix warning
_index.js:2178 Warning: Stateless function components cannot be given refs. Attempts to access this ref will fail._
- *material-ui* _TextField_ does not support _ref_ references
- move search box reference to parent div
- fix naming in _focusSearchField_ func. _searchField_ is referring to _searchBox_ then trying to find _searchField_


varl:
LGTM

PR #224, opened 2018-05-15
Interpretations panel

tokland:
Fixes DHIS2-3428. @turban  @adrianq

Some notes:

- Previously I opened a PR against maps-app (dhis2/maps-app/pull/10), but it should be a shared component, so I've moved here to `packages/interpretations`, now with tests.

- To see the component working, you may use this demo app: https://github.com/EyeSeeTea/d2-ui-playground/tree/feature/interpretations

- The views are taken from interpretations-app. Once Joe Cooper (joe@dhis2.org) closes the UX, we'll update the views/tests. https://invis.io/6FIB6XFDN5R#/286131624_Details_-_Interpretations

- On the implementation: the component can be either controlled or uncontrolled. Usually you'd mount it controlled, since we need to pass/detect changes of the current interpretation.

screenshot from 2018-05-15 14-01-08


varl:
Hi @tokland.

Good work. This component needs an example in the `examples/create-react-app` app to showcase how it is supposed to be used. Can you add it?

You can use Lerna to add deps to `examples/create-react-app` without publishing to NPM with: `lerna add --scope example-cra @dhis2/d2-ui-interpretations`

tokland:
Hi @varl. Indeed, an example was missing, added. (Thanks for the tip on Lerna, it worked like a charm)

varl:
Btw, next time, could you create the branch on our origin and push that to code review? It makes it a lot easier for me to test the code.

If it's an access rights thing, please let me know and I can open d2-ui for externals.

Thanks.

PR #233, opened 2018-06-08
D2 Container similar to Paper from Material-UI

joaoantunes87:
The idea here is to have a d2 container similar to what material-ui offers with Paper.


varl:
We are going to re-work all the core components in relation to this task: https://jira.dhis2.org/browse/TECH-59, and anything we add into Core now will either be scrapped or moved, so I think it's safest to pull in MUI for now.

PR #234, opened 2018-06-08
Interpretations: remove edit details/sharing + tweaks on component reloading/notification

tokland:
Fixes DHIS2-3428

- Remove edit details/sharing buttons and dialogs
- Add prop lastUpdated to control component reload on demand 
- Notify parent onCurrentInterpretationChange on first load


packages/interpretations/src/components/Interpretations.js
```diff
@@ -5,6 +5,7 @@ import { LoadingMask } from '@dhis2/d2-ui-core';
 import { getFavoriteWithInterpretations } from '../models/helpers';
 import DetailsCard from './details/DetailsCard';
 import InterpretationsCard from './interpretations/InterpretationsCard';
+import _ from 'lodash';
```
varl: This adds the entire lodash library to the bundle. Please explicitly only import what you are using, e.g.:

```
import isEqual from 'lodash/fp/isEqual'
import pick from 'lodash/fp/pick'
```

This lets us use tree-shaking and dead code elimination to reduce the bundle size.

tokland: Good point. Refactor this file and other 3 other with usages of lodash.

varl: Great!

PR #246, opened 2018-06-15
[interpretations] Ckeditor and mentions for interpretations and comments

tokland:
Fixes https://jira.dhis2.org/browse/DHIS2-3422

Requires https://github.com/dhis2/maps-app/pull/13

- Textarea for interpretation and interpretation comments use now CKEditor with link/emoji buttons.
- Rich editor supports mentioning (`@username`)
- It uses same markup than analytics east region, will be updated to new markup on next iteration.

### DIFF
--- a/packages/interpretations/src/components/html-editor/CKEditor.js
+++ b/packages/interpretations/src/components/html-editor/CKEditor.js
@@ -1,0 +1,309 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { noop, some } from 'lodash/fp';
+import log from 'loglevel';
+
+/* Additions from standard CKEditor:
+
+    - Tweak internal editor CSS (CKEDITOR.addCss)
+    - Tweak external (toolbar, dialogs) CSS
+    - Add cke_smiley class to dialog to allow custom styling.
+    - Dock position of smiley dialog to button
+    - Hide dialogs on click outside dialog.
+*/
+
+class WrappedEditor {
+    constructor(editor) {
+        this.editor = editor;
+    }
+
+    getValue() {
+        return this.editor.getData();
+    }
+
+    setCursorAtEnd() {
+        setTimeout(() => {
+            this.editor.focus();
+            const range = this.editor.createRange();
+            const selection = this.editor.getSelection();
+            if (range && selection) {
+                range.moveToElementEditEnd(range.root);
+                this.editor.getSelection().selectRanges([range]);
+            }
+        }, 300);
+    }
+
+    getPosition(offset = null) {
+        const range = this.editor.getSelection().getRanges()[0];
+        const bbox = range ? range.startContainer.$.parentNode.getBoundingClientRect() : {top: 0, left: 0};
+        const iframe = this.editor.container.$.getElementsByTagName("iframe")[0];
+        const iframeBbox = iframe.getBoundingClientRect();
+        const pos = { top: bbox.top + iframeBbox.top, left: bbox.left + iframeBbox.left };
+        return offset ? { top: pos.top + offset.top, left: pos.left + offset.left } : pos;
+    }
+
+    getCurrentWord() {
+        const range = this.editor.getSelection().getRanges()[0];
+        if (range) {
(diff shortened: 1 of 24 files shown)


packages/interpretations/src/components/html-editor/CKEditor.js
```diff
@@ -0,0 +1,309 @@
...
+        const iframe = this.editor.container.$.getElementsByTagName("iframe")[0];
+        const iframeBbox = iframe.getBoundingClientRect();
+        const pos = { top: bbox.top + iframeBbox.top, left: bbox.left + iframeBbox.left };
+        return offset ? { top: pos.top + offset.top, left: pos.left + offset.left } : pos;
+    }
+
+    getCurrentWord() {
+        const range = this.editor.getSelection().getRanges()[0];
+        if (range) {
+            const text = range.startContainer.type === CKEDITOR.NODE_TEXT ? range.startContainer.getText() : "";
```
varl: How is the global injected into the app? Does the app have to do something like: https://github.com/dhis2/maintenance-app/blob/bd75c6855eee603ce24fbd7b31464b6e1071d3c2/webpack.config.js#L124 to make it work?

If so, it needs to be added to the README.

tokland: Exactly, we took maintenance-app as reference: https://github.com/dhis2/maps-app/pull/13/files

Added to README: c466eea

packages/interpretations/src/components/html-editor/RichEditor.js
```diff
@@ -0,0 +1,255 @@
...
+        padding: "5px 15px",
+        borderBottom: "1px solid rgb(224, 224, 224)",
+        fontSize: "0.75rem"
+    },
+    userMentionSelected: {
+        backgroundColor: "#ACD",
+    },
+};
+
+class UserMatch extends Component {
```
varl: Sorry, I just noticed that you have implemented a component for giving suggestions when a user mentions another user.

@edoardo is working on a wrapper component in https://github.com/dhis2/d2-ui/tree/task/mentions-wrapper which we can use to wrap a text field or text area, e.g.

```
<MentionsWrapper d2={d2}>
    <TextField
        multiline
        rowsMax="4"
        placeholder="Type some text. @ triggers the mentions suggestions"
    />
</MentionsWrapper>
```

This is to get a pluggable suggestion component we can reuse instead of implementing it in each application. 

Going forward we are looking into deprecating CkEditor completely as it is a rather large dependency (500+ KB), and we want to let each application be standalone and not depend on e.g. core-resources.

I tried adding the `MentionsWrapper` to the interpretations panel in master, and it mostly works. There are a couple of bugs which @edoardo is working on and the components needs more testing before it's ready to merge, but the plan is to get it in for 2.30.

The requirements here are as far as I understand it:

1. Mentioning users
2. Suggestion box for usernames
3. Emoticons
4. Rich-text editor

Using the `MentionsWrapper` solves 1. and 2., which add a lot of value, so I think that would be a good start. I don't think it's reasonable nor possible to combine the `MentionsWrapper` with the CkEditor however, so I would rather we just use `TextField`/`textarea` + `MentionsWrapper` for this iteration until we figure out what to do with requirements 3 and 4.

varl:
Hi @tokland 

Thanks for being patient. Just so that there is no misunderstanding here: I don't have any qualms about the quality of the code in this PR. I found it impressing that you got mentions working with CkEditor in a clean way. Good work!

After a lot of back and forth the way forward is this: 

1. The highest priority is to have a consistent UI when it comes to text input across all apps so I am going to merge this PR into master. 

2. @edoardo will put the mentions component he was implementing on-hold and extract the component you have created for CkEditor w/ Mentions into another package that he can use in Dashboards app.

3. We will create a migration plan for moving away from CkEditor in the future as a whole, and update all the apps to the new solution and migrate the HTML to the new format we are planning to use.

tokland:
@varl I understand, no problem at all.

2. Cool! In this second iteration, we should probably smooth some rough edges regarding the CKEditor integration, for example [https://github.com/dhis2/d2-ui/pull/246/files#diff-67b23b44f9b10f093e7f4fc0bb49bdfeR24](setCursorAtEnd).

3. Good to hear, CKEditor is quite a beast!

Issue #252, opened 2018-06-25
Period Type does not contemplate BiWeekly Period Type


joaoantunes87:
While working on the new Reports app I have noticed that Period Type component does not accept the BiWeekly Period Type, one of the types returned from /api/periodTypes.

joaoantunes87:
A proposal to fix this was create: [https://github.com/dhis2/d2-ui/pull/253](https://github.com/dhis2/d2-ui/pull/253).

PR #253, opened 2018-06-25
Bi Weekly Period

joaoantunes87:
This is a proposal to fix: [https://github.com/dhis2/d2-ui/issues/252](https://github.com/dhis2/d2-ui/issues/252).

- Adding support for BiWeekly Period Type for Period Picker;

Notes:
- I would need some help on making sure it does not breaks anything;
- It also seems to need sync on d2 lib on parsePeriod function at d2/lib/period/parser;

### DIFF
--- a/packages/core/src/period-picker/PeriodPicker.component.js
+++ b/packages/core/src/period-picker/PeriodPicker.component.js
@@ -11,6 +11,7 @@ const styles = {
     year: { width: 95, marginRight: 16 },
     month: { width: 125 },
     week: { width: 105 },
+    biWeek: { width: 105 },
     biMonth: { width: 200 },
     quarter: { width: 200 },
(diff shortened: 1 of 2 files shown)


packages/core/src/period-picker/PeriodPicker.component.js
```diff
@@ -178,6 +181,18 @@ class PeriodPicker extends React.Component {
         return this.renderOptionPicker('week', weeks);
     }
 
+    renderBiWeekPicker() {
```
varl: There are a lot of complexities related to calendar time, so some of the assumptions here do not hold up: e.g. some years have 53 weeks. Every 28 years there is 54 weeks.

We should introduce Moment JS to deal with the complexity related to time and date.

varl:
This has been stale for a long time. Closing.

PR #254, opened 2018-06-25
Fix/ckeditor example app

tokland:
Fixes d2-ui-interpretations in example app.

### DIFF
--- a/examples/create-react-app/src/index.js
+++ b/examples/create-react-app/src/index.js
@@ -6,6 +6,14 @@ import registerServiceWorker from './registerServiceWorker';
 
 import { init } from 'd2/lib/d2'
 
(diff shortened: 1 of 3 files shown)


examples/create-react-app/src/index.js
```diff
@@ -15,7 +23,8 @@ if (!baseUrl) {
 init({
     baseUrl: baseUrl + '/api/29'
 }).then(d2 => {
-    console.log(d2)
+    console.log(d2);
```
edoardo: Perhaps remove this log

edoardo:
Actually, can you open a PR on the d2-ui repo directly?

examples/create-react-app/src/index.js
tokland: Done. I have to admit I tend also to add those `console.log(d2)` / `window.d2 = d2` to work with d2 directly from the console.

PR #257, opened 2018-06-28
[interpretations] Use action Save&share on interpretation create and Share on existing

tokland:
Fixes DHIS2-3428

- On interpretation create: show action "Save & share". It saves the interpretation and then opens the sharing dialog of this new interpretation. Note that the interpretation starts with the same sharing the parent object has.
- On interpretation edit: action "Save" (and does not open sharing dialog)
- Add action "Share" to interpretation on detail.

screenshot from 2018-06-28 14-18-36
screenshot from 2018-06-28 14-19-01


turban:
Could you add both "SAVE & SHARE" and a "SAVE" button when adding new interpretations?

CANCEL . SAVE & SHARE . SAVE

PR #259, opened 2018-07-02
[interpretations] Remove ckeditor and mentions from interpretations/comments

tokland:

- Ckeditor removed, TextField (interpretations) and textarea (comments) restored
- Reply is kept: insert mention to user and focus textarea.
- example app updated


turban:
This PR looks fine to me. We still need to check if the mentioning component from @edoardo can be used for 2.30.

packages/interpretations/src/components/interpretations/CommentTextarea.js
```diff
@@ -3,31 +3,42 @@ import PropTypes from 'prop-types'
...
         onPost: PropTypes.func.isRequired,
         onCancel: PropTypes.func,
-        mentions: PropTypes.object,
     };
 
     constructor(props) {
         super(props);
-        this.state = { text: props.comment.text || "", refresh: new Date() };
+        this.state = { text: props.comment.text || "" };
         this.onPost = this.onPost.bind(this);
```
john-melin: Should probably remove the bind functions and convert to class properties instead to avoid boilerplate code. 

State can also be moved out of the constructor so that the entire constructor can be deleted.

packages/interpretations/src/components/interpretations/CommentTextarea.js
```diff
@@ -3,31 +3,42 @@ import PropTypes from 'prop-types'
 import { Link, ActionSeparator } from './misc';
 import i18n from '@dhis2/d2-i18n'
 import styles from './InterpretationsStyles.js';
-import RichEditor from '../html-editor/RichEditor';
 
 class CommentTextarea extends React.Component {
     static propTypes = {
         comment: PropTypes.object.isRequired,
```
john-melin: Could move out the proptypes to the bottom of the file to follow convention.

packages/interpretations/src/components/interpretations/InterpretationDialog.js
```diff
@@ -6,56 +6,50 @@ import defer from 'lodash/fp/defer';
...
-        this.cancel = this.hideEditorAndThen(this._cancel.bind(this));
-    }
-
-    hideEditorAndThen(fn) {
-        // Method componentWillUnmount of child components of Dialog are called *after* their nodes
-        // are removed from the DOM (bug in mui?), which triggers errors in CKEditor.destroy.
-        // Workaround: Manually unmount the component using a flag in state.
-        return (...args) =>
-            this.setState({ showEditor: false }, () => defer(() => fn(...args)));
+        this.save = this.save.bind(this);
```
john-melin: Same as in the CommentTextarea. Change functions to class properties to avoid using constructor.

tokland:
Merged with latest changes in master.

PR #282, opened 2018-08-27
DHIS2-4079: first implementation of rich text editor and parser

edoardo:
Fixes DHIS2-4079.

- parser: the implementation uses MarkdownIt library, where all the default rules
are disabled.
Instead a couple of custom rules are implemented end enabled.
The supported syntax is italic, bold and a few emojis.
The linkify feature is enabled, which automatically converts URLs or
email addresses into links ( tags).
- editor: simple rich text editor for italic and bold, only keyboard shortcuts for now, no UI)


linxd:
why not using html parser directly?

PR #283, opened 2018-08-27
Upgrade material-ui for the favorites-menu package

jenniferarnesen:


varl:
note to self, maybe we should just delete the favorites source code?

varl:
Looks good to me, approved.

PR #287, opened 2018-09-03
task/update onClick function to remove console warnings

jenniferarnesen:
The console warning reports that the onClick function cannot resolve to false, which is what happens with statements like this:

{ someCondition && returnVal }


jenniferarnesen:
Fixes made in a separate PR

PR #289, opened 2018-09-07
task/Upgrade translation dialog to material-ui 1.5

jenniferarnesen:
Fixes DHIS2-4639

Breaking change:
- upgrade to material-ui 1.5.1
- limit properties that are spread to the mui Dialog component
- fixed console warnings
- fixed error where switching languages caused controls to not get data correctly set (related to the React warnings)
- updated tests


edoardo:
One thing I noticed is that keyboard navigation of the language list is not working anymore.
I used to be able to jump in the list to the first language starting with the letter I type.

edoardo:
approved

PR #292, opened 2018-09-26
Enable support for initially selected periods in period-selector-component

neeilya:
As title says enable support for initially selected periods.
Typical use case: periods selected in favorite/Analytic object.

### DIFF
--- a/examples/create-react-app/src/components/period-selector-dialog.js
+++ b/examples/create-react-app/src/components/period-selector-dialog.js
@@ -8,6 +8,12 @@ import PeriodSelectorDialog from '@dhis2/d2-ui-period-selector-dialog';
 
 class PeriodSelectorDialogExample extends Component {
     state = {
+        // example of initially selected periods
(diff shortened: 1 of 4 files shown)


examples/create-react-app/src/components/period-selector-dialog.js
```diff
@@ -8,6 +8,12 @@ import PeriodSelectorDialog from '@dhis2/d2-ui-period-selector-dialog';
 
 class PeriodSelectorDialogExample extends Component {
     state = {
+        // example of initially selected periods
+        periods: [
+            { id: 'TODAY', name: 'Today' },
+            { id: 'YESTERDAY', name: 'Yesterday' },
+            { id: 'LAST_3_DAYS', name: 'Last 3 days' },
```
edoardo: Does the name (which seems to be the translated string used in a label) need to be passed via props?
I expect typically the app to only pass the list of ids

PR #295, opened 2018-10-02
Add forceReload parameter to children.load call so they are actually re-fetched

HendrikThePendric:
**UPDATE**: Please ignore this for now. It seems this fix just masks an underlying problem.

Fixes DHIS2-966

This doesn't fully fix DHIS2-966 but will do so partly. At least the UI is updated after an OrgUnit is moved

- Add a second `forceReload` parameter with value `true` to the call to `root.children.load`.
- By adding this parameter d2 will actually fetch new data, instead of just returning the existing collection. See the [relevant code in d2 here](https://github.com/dhis2/d2/blob/master/src/model/ModelCollectionProperty.js#L165-L167).

### DIFF
--- a/packages/org-unit-tree/src/OrgUnitTree.component.js
+++ b/packages/org-unit-tree/src/OrgUnitTree.component.js
@@ -102,7 +102,9 @@ class OrgUnitTree extends React.Component {
             this.setState({ loading: true });
 
             const root = this.props.root;
-            root.children.load({ fields: 'id,displayName,children::isNotEmpty,path,parent' }).then((children) => {
+            // d2.ModelCollectionProperty.load takes a second parameter `forceReload` and will just return
+            // the current valueMap unless either `this.hasUnloadedData` or `forceReload` are true
+            root.children.load({ fields: 'id,displayName,children::isNotEmpty,path,parent' }, true).then((children) => {
                 this.setChildState(children);
             });
         }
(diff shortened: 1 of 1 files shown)


HendrikThePendric:
@varl: I re-opened this one because I can't fix DHIS2-966 without it. 

Looking at the code, I don't think this will lead to any unnecessary GET requests in other instances of the OrgUnitTree, because the call to `root.children.reload` is already wrapped in a condition that prevents unnecessary requests.

HendrikThePendric:
Same logic as before, but now added as prop `foreceReloadChildren ` so it will only affect the behavior OrgUnitTrees with this property switched to true.

packages/org-unit-tree/src/OrgUnitTree.component.js
```diff
@@ -369,6 +381,7 @@ OrgUnitTree.defaultProps = {
     hideCheckboxes: false,
     hideMemberCount: false,
     orgUnitsPathsToInclude: null,
+    foreceReloadChildren: false,
```
neeilya: @HendrikThePendric 
a little typo in `foreceReloadChildren` here and above 😉

HendrikThePendric: Oh shoot. Thanks for pointing that out. Will fix.

PR #303, opened 2018-10-17
[d2-ui-interpretations] Update to material-ui v3

tokland:
Reference: https://jira.dhis2.org/browse/DHIS2-4648

- Upgrade material-ui from 0.19 to 3.2.2 (latest)
- Create our custom CollapsibleCard (mui v3 removed the expandable feature from cards)
- Use `mount` instead of `shallow` on some components that needs testing on nested components.
- Version bumped to 1.2.1

Some notes:

- When testing with maps-app, I had a warning: `warning.js:34 Warning: Material-UI: you are using the deprecated typography variants that will be removed in the next major release.`. I fixed it by adding `useNextVariants: true` in `./src/constants/dhis2.theme.js`:

```
export const muiTheme = createMuiTheme({
   colors,
   palette,
   typography: {
       fontFamily: 'Roboto, Helvetica, Arial, sans-serif',
       useNextVariants: true,
   },
 ....
```
- In `examples/create-react-app` I had to remove remove some references in App.js so the page would load: `PeriodSelector`,  `PeriodSelectorDialog`,  `RichTextExample`.  I guess those packages are missing in package.json.

screenshot from 2018-10-17 10-45-58


amcgee:
Hi @tokland, thanks for putting this together.  A few follow-ups:

1) I get the following error multiple times and a Blank Screen of Death when opening the InterpretationsDialog in the maps app (after linking the updated version).  How have you been setting up to test in the maps app?

screenshot 2018-10-18 02 16 13

2) The Interpretations panel does successfully render in the Create React App example, but it's actually hidden because of its positioning in that app.  Can you update the example app so that we can test and validate changes to the component there?  It's ending up behind the Tree examples above it.

3) There are several references to `muiTheme` in component context or childContext which should be removed.  This context object is no longer used in MUI.  In particular `CollapsibleCard:107` logs an error in development mode because the PropType check fails, and there is no need for that context reference in that component.  Other instances are in tests: `DetailsCard.spec.js`, `CommentTextArea.spec.js`, `InterpretationsDialog.spec.js`, and `InterpretationsCard.spec.js` all shouldn't need to manually specify muiTheme in their contexts.

4) (just a note) We use an almost-identical collapsible card in the left-hand sidebar of the Maps app.  Ideally we should refactor this so that we have a canonical collapsible card component and use it in both cases, but that's probably beyond the scope of this PR.

Thanks!
Austin

tokland:
Hi @amcgee,

1. I test with maps-app (just with a minor change so it directly loads a map and interpretation) with a `yarn link "@dhis2/d2-ui-interpretations"`.. I got this error about @global on my first try. I removed node_modules in maps_app, re-installed, and never appeared again.

2. I added a div clear:both after the tree div.

3. I removed the reference in CollapsibleCard, indeed it's not needed there. Also removed muiTheme from childContextTypes in tests. Kept for the MuiThemeProvider (otherwise I get `Material-UI: you are using the deprecated typography variants`).

4. ok!

amcgee:
Thanks @tokland looks good, checking locally now.

amcgee:
> 3\. I removed the reference in CollapsibleCard, indeed it's not needed there. Also removed muiTheme from childContextTypes in tests. Kept for the MuiThemeProvider (otherwise I get `Material-UI: you are using the deprecated typography variants`).

Yeah, the `<MuiThemeProvider theme={muiTheme} />` component is fine, it just doesn't inject the `muiTheme` property into the React context object anymore.  So your change to remove the context but leave the provider is perfect, thanks!

tokland:
@amcgee cool, thx for the review!

PR #307, opened 2018-10-17
Extract selector logic into separate component

neeilya:

- Extract org units selection into separate component


jenniferarnesen:
Looks good. Just one thing - the publish instructions in README are wrong, since we are (currently) supposed to use Lerna for publishing.  You could instead add a link to the RELEASING.md page

PR #318, opened 2018-10-29
BREAKING: Remove UMD bundles, move d2-ui-app to d2-ui-core, etc.

varl:
Before releasing the next major version of d2-ui I want to do some larger changes.

# Breaking changes
- move d2-ui-app to d2-ui-core as it belongs with the 0.20 components
- remove UMD builds on the basis that libraries should only be transpiled, not bundled, since they are included in an application which will run a bundle step and tree-shake at that point
- release d2-ui under a dependent version scheme (https://github.com/lerna/lerna#fixedlocked-mode-default)
- combines dialogs under `@dhis2/d2-ui-dialogs` and editors under `@dhis2/d2-ui-editors`

# CI changes
- introduce deploy-build to d2-ui

# Minor fixes
- remove useless log statements (e.g. in headerbar)
- fix warnings in example app
- remove hacky resolution in package.json as it seems to work without it now


varl:
### rationale about breaking changes

> move d2-ui-app to d2-ui-core as it belongs with the 0.20 components

This component should never have been pulled out to a standalone package in the first place as it is way too small and is only useful together with 0.20 of MUI.

All components which need `App` are pulling in `core` anyway, so makes sense to keep it there.

> remove UMD builds on the basis that libraries should only be transpiled, not bundled, since they are included in an application which will run a bundle step and tree-shake at that point

A rule-of-thumb is that:

**Libraries** should only be transpiled, and **Apps** should be transpiled _and_ bundled.

My thinking is that:
- **Apps** should use the ES modules for integration.
- **Apps** do the final optimizations for production.

I don't seen a lot of arguments for keeping the UMD bundles as they increase the build times from 25 seconds to 240 seconds, and we don't have any tests to catch errors in the bundles (headerbar bundle has been broken for 6 months and no one has noticed), and no applications of ours are using the UMD bundled d2-ui components.

varl:
# Versioning

As a lot of **d2-ui** depends on **d2** we have two options:

1. release **d2-ui** without tying the version to the DHIS2 API (e.g. `v5.0.0`)
2. release **d2-ui** and tying it to the version of the DHIS2 API (e.g. `v31.0.0`)

# Rationale

>1. release **d2-ui** without tying the version to the DHIS2 API (e.g. `v5.0.0`)

It would be very nice to be able to use **d2-ui** without thinking about which API it is pointed to, and reasonable to expect from a UI library.

We could add the **d2** dependency as a _real_ dependency (as opposed to a _peerDep_) to allow an application to install **d2-ui** without thinking about which version of **d2** it needs.

This triggers an immediate need to rewrite several components from accepting **d2** as a _prop_, to be able to bootstrap a full **d2** instance themselves.

One thing to note is that we have historically had problems different **d2** versions used in the **app** and in **d2-ui**. At that point this conflict was resolved using resolve hacks in webpack and `package.json`, which is something I do not want to return to (and this PR actively prohibits by removing UMD bundles).

>2. release **d2-ui** and tying it to the version of the DHIS2 API (e.g. `v31.0.0`)

Since **d2** is often defined as a peer dependency, the **app** which uses **d2-ui** has to make sure that it is using a compliant **d2** package.

Even if we get a warning about an incorrect peerDep when running `yarn install`, it's easy to overlook, but if your `package.json` has a list of deps all pointing to the same version ... the outlier would stand out like a sore thumb during PRs and day-to-day ops.

examples/create-react-app/src/components/period-selector-dialog.js
```diff
@@ -4,7 +4,7 @@ import MuiThemeProvider from 'material-ui/styles/MuiThemeProvider';
 import RaisedButton from 'material-ui/RaisedButton';
 import Snackbar from 'material-ui/Snackbar';
 import PropTypes from 'prop-types';
-import PeriodSelectorDialog from '@dhis2/d2-ui-period-selector-dialog';
+import { PeriodSelectorDialog } from '@dhis2/d2-ui-dialogs';
```
jenniferarnesen: Does importing with this syntax mean that we get all the dialogs in our app bundle?

varl:
this is split into:

https://github.com/dhis2/d2-ui/pull/319
https://github.com/dhis2/d2-ui/pull/321
https://github.com/dhis2/d2-ui/pull/322
https://github.com/dhis2/d2-ui/pull/320

examples/create-react-app/src/components/period-selector-dialog.js
varl: No, it's still shaken out.

PR #321, opened 2018-10-30
Use deploy build and inline app into core

varl:
- move app into core
- introduce deploy-build
- use dependent versioning strategy

> move d2-ui-app to d2-ui-core as it belongs with the 0.20 components

This component should never have been pulled out to a standalone package in the first place as it is way too small and is only useful together with 0.20 of MUI.

All components which need `App` are pulling in `core` anyway, so makes sense to keep it there.

# Versioning

As a lot of **d2-ui** depends on **d2** we have two options:

1. release **d2-ui** without tying the version to the DHIS2 API (e.g. `v5.0.0`)
2. release **d2-ui** and tying it to the version of the DHIS2 API (e.g. `v31.0.0`)

_Note: Both strategies allow us to automate publishing to NPM._

# Thoughts about versioning strategies

>1. release **d2-ui** without tying the version to the DHIS2 API (e.g. `v5.0.0`)

It would be very nice to be able to use **d2-ui** without thinking about which API it is pointed to, and reasonable to expect from a UI library.

We could add the **d2** dependency as a _real_ dependency (as opposed to a _peerDep_) to allow an application to install **d2-ui** without thinking about which version of **d2** it needs.

This triggers an immediate need to rewrite several components from accepting **d2** as a _prop_, to be able to bootstrap a full **d2** instance themselves.

One thing to note is that we have historically had problems different **d2** versions used in the **app** and in **d2-ui**. At that point this conflict was resolved using resolve hacks in webpack and `package.json`, which is something I do not want to return to (and this PR actively prohibits by removing UMD bundles).

>2. release **d2-ui** and tying it to the version of the DHIS2 API (e.g. `v31.0.0`)

Since **d2** is often defined as a peer dependency, the **app** which uses **d2-ui** has to make sure that it is using a compliant **d2** package.

Even if we get a warning about an incorrect peerDep when running `yarn install`, it's easy to overlook, but if your `package.json` has a list of deps all pointing to the same version ... the outlier would stand out like a sore thumb during PRs and day-to-day ops.

This strategy is inline with how we do versioning with applications elsewhere. It's not a strategy I think is valid long-term, but for now the pragmatic alignment would be beneficial.


jenniferarnesen:
On the side of keeping the version number independent of d2/api version, it means we can still adhere to a proper semver pattern. There might be other reasons than d2 version that cause us to need to do a major version bump.

jenniferarnesen:
In a way, I lean towards having components have a direct dependency on d2, especially if they have to directly handle the response and make decisions based on what is in the response.

jenniferarnesen:
Looks like lerna.json has already been updated away from independent versioning to v31? Also, the documentation (e.g. RELEASING.md) will need to be updated.

PR #322, opened 2018-10-30
Group similar components in packages

varl:
> combines dialogs under `@dhis2/d2-ui-dialogs` and editors under `@dhis2/d2-ui-editors`

There are two distinct groups which are emerging, **editors** and **dialogs**, so I think they should be gathered into respective packages. This reduces the amount of packages we have to publish and maintain.

- editors _(these are all MUI v0.20)_
- - expression-manager
- - formula-editor
- - group-editor
- - rich-text

- dialogs _(these are all MUI v3)_
- - favorites-dialog
- - sharing-dialog
- - period-selector-dialog
- - translation-dialog
- - org-unit-dialog

That leaves the following packages which I have trouble finding groups for:

- forms _(MUI v0.20)_
- icon-picker _(MUI v0.20)_
- legend _(MUI v0.20)_
- org-unit-select _(MUI v0.20)_
- table _(MUI v0.20)_
- file-menu _(MUI v3)_
- org-unit-tree _(MUI v3)_
- interpretations _(MUI v3)_
- mentions-wrapper _(MUI v3)_

Core is a group by itself and header-bar doesn't need a home, it's a root-level package as well.

- core _(MUI v0.20)_
- header-bar _(MUI v3)_


jenniferarnesen:
I generally like the idea of grouping things together, but I'm still not sure what we gain by making the suggested groups. If we are moving back to a single version, and we are going to use travis for deploying, does it matter how many packages we deploy?

Also, rich-text does not depend on MUI at all.

varl:
>If we are moving back to a single version, and we are going to use travis for deploying, does it matter how many packages we deploy?

Not technically, it's more the mental overhead when debugging nested dependencies. I've found errors these last couple of days I have no idea how they even happen. Human error or Lerna error?

>What is the argument for not putting everything back into a single package? Or two: one for -old-mui, and the other for everything else (as suggested by Austin)?

One introduces the age-old problem of multiple MUI versions in the same package, but that should work better now.

Two would be perfectly fine, I think.

>Side note: rich-text does not depend on MUI at all.

Right, still an editor so I put it there.

amcgee:
@varl @jenniferarnesen Are we moving forward on this or was it closed because we are not?

I'm looking for somewhere to put a Mui 3 theme to adopt the modern DHIS2 identity and it doesn't fit well in any of the existing packages - `core` fits logically but it is still on the old MUI.  If we had `d2-ui-modern` and `d2-ui-legacy` or whatever we want to call them the theme would fit in the `modern` package.

I think I'm still in favor of 2 packages, should we ahead and do that?

jenniferarnesen:
Sort of side-notes:

One other thing I have come across as I try to pull old mui out of data-viz in time for 2.31, is that several of the core components do not depend on MUI at all, while other "components" are not actually components (Action, Store, Theme).

What about removing all old-mui components from core, for the next major d2-ui release? We can just use MUI directly in our apps until the d2-ui/ui library is ready.

PR #324, opened 2018-10-31
task/remove packages not in use

jenniferarnesen:
yarn remove dependencies that are not used by the package


varl:
No-brainer :+1:

PR #326, opened 2018-11-01
Task/remove mui beta version from several components

jenniferarnesen:
Mui < v 1.0 has been removed from the following components:

* translation-dialog
* sharing-dialog
* org-unit-tree

Solutions:
- replace d2-ui-core imports (that in turn imported from material-ui) with @material-ui/core
- use import syntax for d2-ui-core that results in tree-shaking

I've done a lot of comparisons of bundle size today (using tools source-map-explorer and whybundled) and I believe that these changes will result in old material-ui disappearing from the data-visualizer bundle.  I did my bundle-size comparisons with the example app, which was a little complicated because the example app itself imports a bunch of old material-ui components. But I chose example app instead of data-visualizer app since sometimes the yarn linking causes unexpected behavior.

I know that the import syntax for d2-ui-core is not pretty, but refactoring d2-ui-core to have a nicer syntax was too big of a job for right now, and with core eventually being replaced by the new ui lib, I wasn't sure the effort would ever be worth it.

### DIFF
--- a/packages/org-unit-tree/src/OrgUnitTree.component.js
+++ b/packages/org-unit-tree/src/OrgUnitTree.component.js
@@ -8,7 +8,7 @@ import StopIcon from '@material-ui/icons/Stop';
 import ModelBase from 'd2/model/Model';
 import ModelCollection from 'd2/model/ModelCollection';
 
-import { TreeView } from '@dhis2/d2-ui-core';
+import TreeView from '@dhis2/d2-ui-core/build/es/tree-view/TreeView.component';
 
 
 const styles = {
--- a/packages/sharing-dialog/src/Sharing.component.js
+++ b/packages/sharing-dialog/src/Sharing.component.js
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import Divider from '@material-ui/core/Divider';
 import Typography from '@material-ui/core/Typography';
-import { Heading } from '@dhis2/d2-ui-core';
+import Heading from '@dhis2/d2-ui-core/build/es/headings/Heading.component';
 import UserSearch from './UserSearch.component';
 import CreatedBy from './CreatedBy.component';
 import {
--- a/packages/sharing-dialog/src/SharingDialog.component.js
+++ b/packages/sharing-dialog/src/SharingDialog.component.js
@@ -14,14 +14,14 @@ const defaultState = {
     errorMessage: '',
 };
 
-const loadingStatusMask = {
+const loadingMaskStyle = {
     position: 'relative',
     left: '45%',
     top: '45%',
 };
 
 const LoadingMask = () => (
-    <div style={loadingStatusMask} >
+    <div style={loadingMaskStyle} >
         <CircularProgress />
     </div>
 );
--- a/packages/sharing-dialog/src/SharingDialog.component.js
+++ b/packages/sharing-dialog/src/SharingDialog.component.js
@@ -5,20 +5,26 @@ import DialogContent from '@material-ui/core/DialogContent';
 import DialogActions from '@material-ui/core/DialogActions';
 import Button from '@material-ui/core/Button';
 import Snackbar from '@material-ui/core/Snackbar';
+import CircularProgress from '@material-ui/core/CircularProgress';
 import PropTypes from 'prop-types';
 import Sharing from './Sharing.component';
-import { LoadingMask } from '@dhis2/d2-ui-core';
-
-const styles = {
-    loadingMask: {
-        position: 'relative',
-    },
-};
 
 const defaultState = {
     sharedObject: null,
     errorMessage: '',
 };
+
+const loadingStatusMask = {
+    position: 'relative',
+    left: '45%',
+    top: '45%',
+};
+
+const LoadingMask = () => (
+    <div style={loadingStatusMask} >
+        <CircularProgress />
+    </div>
+);
 
 /**
  * A pop-up dialog for changing sharing preferences for a sharable object.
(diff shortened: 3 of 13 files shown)


packages/sharing-dialog/src/SharingDialog.component.js
```diff
@@ -5,21 +5,27 @@ import DialogContent from '@material-ui/core/DialogContent';
...
 };
 
+const loadingMaskStyle = {
+    position: 'relative',
+    left: '45%',
+    top: '45%',
+};
+
+const LoadingMask = () => (
+    <div style={loadingMaskStyle} >
```
amcgee: Do we prefer `style` or `className` here?  Usually I would prefer withStyles() and className but I don't know if we've standardized on that.

packages/org-unit-tree/src/OrgUnitTree.component.js
```diff
@@ -8,7 +8,7 @@ import StopIcon from '@material-ui/icons/Stop';
 import ModelBase from 'd2/model/Model';
 import ModelCollection from 'd2/model/ModelCollection';
 
-import { TreeView } from '@dhis2/d2-ui-core';
+import TreeView from '@dhis2/d2-ui-core/build/es/tree-view/TreeView.component';
```
amcgee: This style of import is necessary to enable tree shaking and exclude `material-ui@0.20`?  Does it actually work or is `d2-ui-core` pulling in `material-ui` anyway since it's a dependency in `package.json`?

jenniferarnesen: When I removed all the imports in example app, except for, say, SharingDialog, then the only material-ui that was in the bundle was CircularProgress (this was when I was still importing that from d2-ui-core). So I'm pretty sure that if there is no import statement, then it doesn't get into the bundle, even though it is a dependency.

jenniferarnesen: As far as the import syntax, it may be that it treeshakes even with the import syntax like 
`import { TreeView } from '@dhis2/d2-ui-core';`

But after reading this article, I decided to be on the safe side (example with lodash-es):  https://advancedweb.hu/2017/02/07/treeshaking/

packages/sharing-dialog/src/SharingDialog.component.js
jenniferarnesen: As far as I know, we have not standardized, though I agree with your preferences.

packages/org-unit-tree/src/OrgUnitTree.component.js
jenniferarnesen: Here is an example you can run:  https://github.com/jenniferarnesen/treeshake-test.git

The import syntax matters, big-time.

varl: I added this to `treeshake-test/node_modules/@dhis2/d2-ui-core/package.json`:

```
"sideEffects": [
    "*.css"
],
```

and with that change i get:

```
import Heading from '@dhis2/d2-ui-core/build/es/headings/Heading.component';
// -rw------- 1 varl varl 117K Nov  2 09:41 1.9c28a522.chunk.js

import { Heading } from '@dhis2/d2-ui-core';
// -rw------- 1 varl varl 117K Nov  2 09:42 1.7a19d5fa.chunk.js

import { Heading } from '@dhis2/d2-ui-core/build/es/index.js';
// -rw------- 1 varl varl 117K Nov  2 09:43 1.7a19d5fa.chunk.js
```

I included `@dhis2/d2-ui-core/build/es/index.js` to be absolutely sure that webpack was using the ES-bundle in the `@dhis2/d2-ui-core` case. The bundle has the same hash but different timestamps so I'm confident it's the same.

Without the `sideEffects` prop in `d2-ui-core` I get, just like the results in treeshake-test.git.

```
import Heading from '@dhis2/d2-ui-core/build/es/headings/Heading.component';
// -rw------- 1 varl varl 117K Nov  2 09:46 1.1424eb58.chunk.js

import { Heading } from '@dhis2/d2-ui-core';
// -rw------- 1 varl varl 736K Nov  2 09:47 1.120157cb.chunk.js

import { Heading } from '@dhis2/d2-ui-core/build/es/index.js';
// -rw------- 1 varl varl 736K Nov  2 09:48 1.120157cb.chunk.js
```

I think we should add the `sideEffects` prop in d2-ui components/packages and do some testing. Pretty confident it's fine in d2-ui-core, but the others is more of an open question.

packages/sharing-dialog/src/Sharing.component.js
```diff
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import Divider from '@material-ui/core/Divider';
 import Typography from '@material-ui/core/Typography';
-import { Heading } from '@dhis2/d2-ui-core';
+import Heading from '@dhis2/d2-ui-core/build/es/headings/Heading.component';
```
varl: After flatpacking in #327 this will be: 

```
import Heading from '@dhis2/d2-ui-core/es/headings/Heading.component';
```

If we drop CJS we can make it:

```
import Heading from '@dhis2/d2-ui-core/headings/Heading.component';
```

varl: But we can merge this first and I'll fix my PR.

packages/org-unit-tree/src/OrgUnitTree.component.js
amcgee: Great test cases, thanks both!  I think simplifying the import syntax will be a solid win for DX.

> but we also need to test that our code is actually side-effect free.
:+1: This is key.  If we're doing ES modules correctly it should always be side-effect free, but we might not be.

varl:
Proceed with merge, then remove CJS, and flatten so that ES builds straight to build 👍

packages/org-unit-tree/src/OrgUnitTree.component.js
varl: >👍 This is key. If we're doing ES modules correctly it should always be side-effect free, but we might not be.

Agreed, although some libraries which use imports to extend runtime objects (rxjs i'm looking at you) do trigger side-effects themselves. We'll need to do some deeper testing to see the impacts when we start to claim to Webpack that we don't have side effects. 😨

amcgee: Can't wait!  😨

PR #327, opened 2018-11-01
Bye-bye lerna, thanks for all the fish.

varl:
Ok, so turns out we can get by with just Yarn Workspaces to handle the inter-package deps, a simple runner to build all packages, and running the tests with jest.

Deployment & publishing is handed to `deploy-builds` which puts the build artifacts for commits, and if a tag is pushed to the repo, it builds and deploys that too, with the addition that if it builds alright it also publishes the packages to NPM.

Some examples of the deployed packages:

- https://github.com/d2-ci/d2-ui-core/blob/bye-bye-lerna/package.json
- https://github.com/d2-ci/d2-ui-sharing-dialog/blob/bye-bye-lerna/package.json
- https://github.com/d2-ci/d2-ui-table/blob/master/package.json
- https://github.com/d2-ci/d2-ui-translation-dialog/blob/bye-bye-lerna/package.json
- https://github.com/d2-ci/d2-ui-period-selector-dialog/blob/bye-bye-lerna/package.json


docs/RELEASING.md
```diff
@@ -1,221 +1,30 @@
...
-```
+The release process has been simplified as to be executed by Travis
+only using [deploy-build](https://github.com/dhis2/deploy-build).
 
-Use the `updated` and `diff` commands to get an idea of what Lerna thinks should be updated.
+# Deployment
 
-```
-$ lerna updated     # used by publish to determine changes
+All commits across all branches is built automatically and published to
```
jenniferarnesen: is -> are

docs/RELEASING.md
```diff
@@ -1,221 +1,30 @@
...
-
-Familiarise yourself with [Lerna](http://lernajs.io/) before proceeding.
-
-You will need it installed globally.
-
-Before running the following lerna commands, make sure you have the latest git tags locally
-```
-git fetch --tags
-```
+The release process has been simplified as to be executed by Travis
```
jenniferarnesen: The release process is executed by Travis using [deploy build]...

(can drop the history part)

docs/RELEASING.md
```diff
@@ -1,221 +1,30 @@
...
+on a successful build using the `deploy-build.sh` script.
 
-$ lerna diff        # used internally to diff against the last released tag and
-                    # prints the code changes which is useful for determining semver
-```
+# Publish to NPM
 
-To avoid creating tags, pushing them, etc. only to fail on the actual
-publishing to NPM make sure you are logged in.
+To publish to Github the developer needs only to create a version tag
```
jenniferarnesen: To publish to NPM, create a version tag locally using ...

docs/RELEASING.md
varl: oops.

docs/RELEASING.md
varl: Thanks, will fix.

PR #328, opened 2018-11-02
Task/remove cjs build from all packages

jenniferarnesen:
The commonjs build is not in use and is being removed to simplify the build

Changes include:
- remove cjs build
- remove webpack loaders that hadn't previously been removed from packages
- remove obsolete pre-commit steps
- in period-selector-dialog, put css file in same directory as js files in build/

### DIFF
--- a/packages/sharing-dialog/src/Sharing.component.js
+++ b/packages/sharing-dialog/src/Sharing.component.js
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import Divider from '@material-ui/core/Divider';
 import Typography from '@material-ui/core/Typography';
-import Heading from '@dhis2/d2-ui-core/build/es/headings/Heading.component';
+import Heading from '@dhis2/d2-ui-core/build/headings/Heading.component';
 import UserSearch from './UserSearch.component';
 import CreatedBy from './CreatedBy.component';
 import {
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -29,10 +29,6 @@
     "react": "^16.0.0",
     "react-dom": "^16.0.0"
   },
-  "pre-commit": [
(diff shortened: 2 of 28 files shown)


packages/sharing-dialog/src/Sharing.component.js
```diff
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import Divider from '@material-ui/core/Divider';
 import Typography from '@material-ui/core/Typography';
-import Heading from '@dhis2/d2-ui-core/build/es/headings/Heading.component';
+import Heading from '@dhis2/d2-ui-core/build/headings/Heading.component';
```
varl: I think these imports are going to break in "production", because the `build` folder won't exist, and the `package.json` properties for `main` and `module` will be rewritten to drop the `build/` part when they are copied to the build directory for flatpacking.

varl: I've been thinking some more on this over the weekend and it's not a huge deal to me if we use `/build/` in the docs, but not in production. The docs app for d2-ui is local anyway since it requires a backend so we cannot deploy or use it with production anyway.

We just need to be clear that this:
`import Heading from '@dhis2/d2-ui-core/build/headings/Heading.component';`

Will be:
`import Heading from '@dhis2/d2-ui-core/headings/Heading.component';`

When d2-ui is used from a real app.

We could use `packages` to `link` to the `build/` directory locally as well but that introduces more build steps, so I think it's nicer to rely on Yarn workspaces locally, and publish the flat packages from build and get clean imports.

jenniferarnesen: So how will this work during development in d2-ui, with linked packages? I guess the link will be all the way to the build dir?

PR #329, opened 2018-11-05
task/remove "build" from import path of d2-ui-core components

jenniferarnesen:
When the package is flatpacked, the build directory will not exist.

In D2UIApp.spec.js I removed an import and some variables that are not used.


jenniferarnesen:
The tests will fail until the packages are published.

PR #330, opened 2018-11-06
Add a Mui3 theme configuration to d2-ui/core (no additional MUI dependency)

amcgee:
Fixes DHIS2-5044.

- Add a [dhis2 theme](https://projects.invisionapp.com/share/A7LT4TJYETS#/screens/302550228_Color) configuration compatible with [Material UI version 3 themes](https://material-ui.com/customization/themes/) to `d2-ui/core`

---

**What does this theme do?**

* **Palette** - Set Mui color palette defaults to match dhis2 identity colors
* **Spacing** - Set base unit spacing to 8px (also the Mui default)
* **Typography** - Use Roboto font family and `useNextVariants` (to prevent deprecation warnings)
* **Overrides** - Use the Mui theme override mechanism to create a standardized look and feel for primitive components (Dividers and Dialogs, for now)
** In particular, make dialog actions and titles appear consistently in accordance with this design spec by @joeDHIS 

---

Note that, to avoid adding a dependency on `@material-ui/core@3` to `d2-ui/core` (which currently relies on `material-ui@0.20`) I've omitted `createMuiTheme` and export just the raw theme object.

**USAGE:**

```javascript
import { createMuiTheme, MuiThemeProvider } from '@material-ui/core/styles'
import { mui3theme as dhis2theme } from '@dhis2/core'
// OR: import { theme as dhis2theme } from '@dhis2/core/theme/dhis2.mui3.theme'

const MyApp = (
    <MuiThemeProvider theme={createMuiTheme(dhis2theme)}>
        ...<MyAppComponents />...
    </MuiThemeProvider
)
```

To override parts of the theme (for development purposes only, the canonical theme should live in `d2-ui`):

```javascript
import { theme as dhis2theme } from '@dhis2/d2-ui-core/theme/dhis2.mui3.theme'
const customTheme = {
  ...dhis2theme,
  overrides: {
    ...dhis2theme.overrides,
    MuiToolbar: { ... },
  },
  ...etc
}
```

---

Future improvements:

* Call `createMuiTheme` in the theme file and export the result
* Use `<MuiThemeProvider>` from Mui 3 and this theme in `d2-ui/core/App` instead of or in addition to the Mui 0.20 theme
* Add `<MuiThemeProvider>` using this theme to `examples/create-react-app` (CRA example is currently broken on master, @varl is working on a fix)
* Define the distinction between a **Modal Dialog** with complex content and a **Modal Alert** with just some information and a single button (or whatever we want to call them).  As it stands, after this change all popup Dialogs will have the divider line between Content and Actions.  We may not want that line for simple alerts or prompts, but this hasn't been fully fleshed out yet (@joeDHIS).  Samples of current dialog and prompts with this theme are below.
* Update `d2-ui` Dialog components so they don't customize on top of these default styles (see [here](https://jira.dhis2.org/browse/DHIS2-5043))

---

**Rules for successful app theming and maintanability** (plus easier migration to `@dhis2/ui` some day)

* **Apps should pretty much never use hard-coded colors.**  Always use a color from the theme, or add one to the theme if absolutely necessary
* **Use multiples of `theme.spacing.unit` for margins, padding, and positioning**  Don't hard-code pixel values like `12px`, instead use `theme.spacing.unit * 1.5` - this makes the entire application layout proportionally responsive to theme changes
* **Don't customize individual instances of a component (like Dialog, DialogTitle, Button, etc)** - instead, use the defaults provided by Mui with the overrides provided by this theme, or if necessary update the overrides here so the style changes are consistently applied to all DHIS2 application dialogs rather than just to a single one.
* Use `<Typography>` instead of raw HTML innerText.

---

**Examples**

Full Modal Dialog:

screen shot 2018-11-06 at 11 58 15 am

Simple Modal Alert:

screen shot 2018-11-06 at 11 51 50 am


amcgee:
Note that tests are failing on Master, so they are also failing for this PR.

jenniferarnesen:
Looks good. I think its a good thing that the theme file is pure with no dependencies, and that the apps themselves need to call createMuiTheme.

One note, is the "dhis2" needed in the filename?  Its already namespaced with @dhis2.

packages/core/src/theme/dhis2.mui3.theme.js
```diff
@@ -0,0 +1,158 @@
...
+        light: colors.accentPrimaryLight,
+        lightest: colors.accentPrimaryLightest, // Custom extension, not used by default
+        // contrastText: 'white',
+    },
+    secondary: {
+        main: colors.accentSecondary,
+        dark: colors.accentSecondaryDark,
+        light: colors.accentSecondaryLight,
+        lightest: colors.accentSecondaryLightest, // Custom extension, not used by default
+        // contrastText: 'white',
```
jenniferarnesen: Do we need to keep the commented-out lines?

amcgee: Technically no, I left them in so that we knew they could be set there (it's the only omitted key, white is the default).  Can remove them if you think it's too messy though.

amcgee:
> Looks good. I think its a good thing that the theme file is pure with no dependencies, and that the apps themselves need to call createMuiTheme.
> 
> One note, is the "dhis2" needed in the filename? Its already namespaced with @dhis2.

Not needed, I'll remove it.

PR #333, opened 2018-11-08
Move theme usage to README, update base fontSize

amcgee:
Also order CSS rules alphabetically.


packages/core/README.md
```diff
@@ -5,13 +5,50 @@ interfaces in dhis2 applications.
...
+```sh
+> yarn add @dhis2/d2-ui-core
 ```
 
 ## Usage
 
 Importing components
+
+```js
+import Button from "@dhis2/d2-ui-core/Button";
```
jenniferarnesen: I think its actually `import Button from "@dhis2/d2-ui-core/button/Button`;

This is because I didn't get time to rename files to index.js.  I do have a partially implemented branch for that though.

amcgee: Ok nice catch, I actually just guessed on that syntax since I was updating the README anyway.

amcgee: Fixed

PR #334, opened 2018-11-08
chore: add back versions to packages

varl:
- add a placeholder version to packages
- fix a broken test using node_modules to get the thing it's testing


packages/core/src/app/__tests__/D2UIApp.spec.js
```diff
@@ -3,7 +3,8 @@ import MuiThemeProvider from 'material-ui/styles/MuiThemeProvider';
 import getMuiTheme from 'material-ui/styles/getMuiTheme';
 import { shallow } from 'enzyme';
 import D2UIApp from '../index';
-import {theme} from '@dhis2/d2-ui-core';
+
+import theme from '../../theme/theme';
```
varl: this is what fixes the tests in ci

PR #335, opened 2018-11-09
Use patch range for d2 31.1.x

varl:


amcgee:
This alleviates an error because of a breaking change in d2 31.2.x, we should investigate and fix the root cause but this is helpful for now.

PR #341, opened 2018-11-09
Task/period selector dialog

neeilya:

- Improved styles according to designs/latest discussions
- Use @dhis2/i18n for translations

### DIFF
--- a/packages/period-selector-dialog/css/PeriodSelector.css
+++ b/packages/period-selector-dialog/css/PeriodSelector.css
@@ -7,6 +7,7 @@ button.nav-button {
     padding: 5px 13px 3px;
     border: 1px solid #ECEFF1;
     border-radius: 2px 0 0 2px;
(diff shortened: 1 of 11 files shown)


packages/period-selector-dialog/css/PeriodSelector.css
```diff
@@ -113,6 +112,12 @@ button.nav-button {
 .selector-area {
     position: relative;
     height: 460px;
+    -webkit-touch-callout: none;
+    -webkit-user-select: none;
```
jenniferarnesen: I'm not sure its necessary to include the browser prefixes. Apps including the component should configure their bundler to do the prefixing (see webpack.config.prod.js in data-visualizer)

neeilya: got it, thanks, will fix now

PR #344, opened 2018-11-20
Task/org unit dialog

neeilya:
Fixes DHIS2-5207.

- Ability to select org unit children using context menu
- Adjusted stylings
    - cursor: pointer for org unit labels & folder icons
    - Align left user org units

### DIFF
--- a/packages/org-unit-dialog/src/OrgUnitDialog.js
+++ b/packages/org-unit-dialog/src/OrgUnitDialog.js
@@ -35,6 +35,7 @@ class OrgUnitDialog extends React.PureComponent {
                     handleOrgUnitClick={this.props.handleOrgUnitClick}
                     handleUserOrgUnitClick={this.props.handleUserOrgUnitClick}
                     checkboxColor={this.props.checkboxColor}
+                    handleMultipleOrgUnitsSelect={this.props.handleMultipleOrgUnitsSelect}
                 />
             </DialogContent>
             <DialogActions>
@@ -115,6 +116,12 @@ OrgUnitDialog.propTypes = {
     * Arguments supplied in callback: event, checked
     */
     handleUserOrgUnitClick: PropTypes.func.isRequired,
+
+    /**
+     * Callback handler for selecting user orgunit
+     * Arguments supplied in callback: event, checked
+     */
+    handleMultipleOrgUnitsSelect: PropTypes.func.isRequired,
 
     /**
     * Root organisation unit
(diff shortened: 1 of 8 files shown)


packages/org-unit-dialog/src/OrgUnitSelector.js
```diff
@@ -70,8 +76,36 @@ class OrgUnitSelector extends Component {
...
+        if (!hasChildren) {
+            return;
+        }
+
+        this.setState({
+            menuAnchorElement: event.currentTarget,
+            loadingChildren: true,
+        });
+
+        loadChildren().then((children) => {
```
jenniferarnesen: Should this statement be a callback to the setStatement above?

packages/org-unit-dialog/src/OrgUnitSelector.js
```diff
@@ -166,13 +200,27 @@ class OrgUnitSelector extends Component {
...
+                            anchorEl={this.state.menuAnchorElement}
+                            open={Boolean(this.state.menuAnchorElement)}
+                            onClose={this.closeContextMenu}
+                        >
+                            <MenuItem
+                                onClick={this.selectChildren}
+                                disabled={this.state.loadingChildren}
+                                dense
+                            >
+                                Select children
```
jenniferarnesen: 'Select children' should be internationalized

packages/org-unit-dialog/src/OrgUnitDialog.js
```diff
@@ -116,6 +117,12 @@ OrgUnitDialog.propTypes = {
     */
     handleUserOrgUnitClick: PropTypes.func.isRequired,
 
+    /**
+     * Callback handler for selecting user orgunit
+     * Arguments supplied in callback: event, checked
+     */
+    handleMultipleOrgUnitsSelect: PropTypes.func.isRequired,
```
jenniferarnesen: So this will be a breaking change then.

jenniferarnesen:
The code seems fine, but OrgUnitTree.component just keeps getting bigger and more complex. Can you spend some time doing some basic refactoring and adding tests? It can be done in a separate PR, but would be good to do this before picking up a new task.

Thoughts on refactoring:

- move styles to a separate file
- pull the loading function into a separate services.js file
- the render function has a bunch of comments that are unnecessary
- the render function has a high (poor) complexity score. Any way to simplify it? (e.g., make a OUCheckbox component, OUFolderIcon component), pull the event handlers out to standalone functions with argument...

packages/org-unit-dialog/src/OrgUnitDialog.js
neeilya: I can make it optional, just wondering if it causes problems?

PR #347, opened 2018-11-22
Task/deselect all orgunits

neeilya:
Fixes DHIS2-5260.

- Enable support for deselecting all org units

### DIFF
--- a/packages/org-unit-dialog/src/OrgUnitSelector.js
+++ b/packages/org-unit-dialog/src/OrgUnitSelector.js
@@ -102,21 +102,6 @@ class OrgUnitSelector extends Component {
 
     renderOptionsPanel = () => (
         <div style={styles.footer.index}>
-            <div style={styles.footer.tooltipContainer}>
-                {this.props.selected.length > 0 && (
-                    <div style={styles.footer.tooltip}>
-                        {this.props.selected.length} selected.
-                        <a
-                            onClick={this.props.onDeselectAllClick}
-                            style={styles.footer.tooltip.link}
-                            // eslint-disable-next-line
-                            href="javascript:void(0)"
-                        >
-                            Deselect all
-                        </a>
-                    </div>
-                )}
-            </div>
             <Grid style={styles.footer.gridContainer} container>
                 <Grid
                     item
@@ -197,6 +182,19 @@ class OrgUnitSelector extends Component {
                             disableSpacer
                         />
                     </div>
+                </div>
+                <div style={styles.orgUnitsContainer.tooltipContainer}>
+                    {this.props.selected.length > 0 && (
+                        <div style={styles.orgUnitsContainer.tooltip}>
+                            {this.props.selected.length} {i18n.t('selected')}.
+                            <button
+                                onClick={this.props.onDeselectAllClick}
+                                style={styles.orgUnitsContainer.tooltip.link}
+                            >
+                                {i18n.t('Deselect all')}
+                            </button>
+                        </div>
+                    )}
                 </div>
             </div>
             <div>{this.renderOptionsPanel()}</div>
--- a/packages/org-unit-dialog/src/OrgUnitSelector.js
+++ b/packages/org-unit-dialog/src/OrgUnitSelector.js
@@ -25,6 +25,7 @@ class OrgUnitSelector extends Component {
         // if props.selected.length changed by more than 1, then another analytic object was selected
         if (Math.abs(prevProps.selected.length - this.props.selected.length) > 1) {
             // In this case refresh expanded org units
+            // eslint-disable-next-line
             this.setState({
                 initiallyExpanded: this.props.selected.map(ou => removeLastPathSegment(ou.path)),
             });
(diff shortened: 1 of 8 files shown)


packages/org-unit-dialog/src/OrgUnitSelector.js
```diff
@@ -43,6 +44,7 @@ class OrgUnitSelector extends Component {
                     counter += 1;
 
                     if (counter > 1) {
+                        // eslint-disable-next-line
```
jenniferarnesen: Is there a danger of infinite updating with setState in CDU?

packages/org-unit-dialog/src/OrgUnitSelector.js
```diff
@@ -99,49 +101,70 @@ class OrgUnitSelector extends Component {
...
+                {this.props.selected.length > 0 && (
+                    <div style={styles.footer.tooltip}>
+                        {this.props.selected.length} selected.
+                        <a
+                            onClick={this.props.onDeselectAllClick}
+                            style={styles.footer.tooltip.link}
+                            // eslint-disable-next-line
+                            href="javascript:void(0)"
+                        >
+                            Deselect all
```
jenniferarnesen: internationalization

packages/org-unit-dialog/src/OrgUnitSelector.js
```diff
@@ -99,49 +101,70 @@ class OrgUnitSelector extends Component {
...
-                        displayEmpty
-                        fullWidth
-                    >
-                        {this.props.groupOptions.map(option => (
-                            <MenuItem key={option.id} value={option.id}>{option.displayName}</MenuItem>
-                        ))}
-                    </Select>
-                </FormControl>
+        <div style={styles.footer.index}>
+            <div style={styles.footer.tooltipContainer}>
```
jenniferarnesen: Since this tooltip isn't an option and its being absolutely positioned, it seems to be logically belong better to the div that wraps the OrgUnitTree. I was also thinking that "tooltip" is a bit misleading as a term. Maybe snackbar or something like that?

packages/org-unit-dialog/src/OrgUnitSelector.js
```diff
@@ -99,49 +101,70 @@ class OrgUnitSelector extends Component {
...
+                            // eslint-disable-next-line
+                            href="javascript:void(0)"
+                        >
+                            Deselect all
+                        </a>
+                    </div>
+                )}
+            </div>
+            <Grid style={styles.footer.gridContainer} container>
+                <Grid
```
jenniferarnesen: Even though this code was not part of this PR, you could remove some code duplication (of the <Grid components) by adding a stateless component like so (not sure if GridControl is the best name):
```
<GridControl
                    title={i18n.t('Level')}
                    id="level-select"
                    value={this.props.level}
                    onChange={this.props.onLevelChange}
                    options={this.props.levelOptions}
                    disabled={this.props.userOrgUnits.length > 0}
                    renderValue={this.renderLevelOptions}
                    multiple
                />```

And 
```
const GridControl = ({id, title, options, ...props}) => {
    return (<Grid
        item
        xs={4}
        style={styles.footer.gridContainer.gridItem}
    >
        <FormControl style={{ width: '100%' }}>
            <InputLabel htmlFor={id}>{title}</InputLabel>
            <Select
                input={<Input id={id} />}
                fullWidth
                {...props}
            >
                {options.map(option => (
                    <MenuItem key={option.id} value={option.id}>{option.displayName}</MenuItem>
                ))}
            </Select>
        </FormControl>
    </Grid>);
}
```

Comment (jenniferarnesen) [reply packages/org-unit-dialog/src/OrgUnitSelector.js]: And the component (might not display correctly here):

``````
const GridControl = ({id, title, options, ...props}) => {
    return (<Grid
        item
        xs={4}
        style={styles.footer.gridContainer.gridItem}
    >
        <FormControl style={{ width: '100%' }}>
            <InputLabel htmlFor={id}>{title}</InputLabel>
            <Select
                input={<Input id={id} />}
                fullWidth
                {...props}
            >
                {options.map(option => (
                    <MenuItem key={option.id} value={option.id}>{option.displayName}</MenuItem>
                ))}
            </Select>
        </FormControl>
    </Grid>);
}```

File: packages/org-unit-dialog/src/OrgUnitSelector.js [hunk 5]
```diff
@@ -99,49 +101,70 @@ class OrgUnitSelector extends Component {
...
-                            <MenuItem key={option.id} value={option.id}>{option.displayName}</MenuItem>
-                        ))}
-                    </Select>
-                </FormControl>
+        <div style={styles.footer.index}>
+            <div style={styles.footer.tooltipContainer}>
+                {this.props.selected.length > 0 && (
+                    <div style={styles.footer.tooltip}>
+                        {this.props.selected.length} selected.
+                        <a
```
jenniferarnesen: Suggest using a styled <button> instead - more semantically correct and no hacking needed.

packages/org-unit-dialog/src/OrgUnitSelector.js
neeilya: Yes, unless "if" condition is implemented.

neeilya:
@jenniferarnesen just pushed changes related to your comments. Thanks.

PR #349, opened 2018-11-22
Fix/periods selector dialog

neeilya:
- Make sure offered periods list does not contain items  Fix a bug where user is not able to select last offered period in periods list

Steps to reproduce the bug:
- Load fresh AO
- Open periods dialog
- Make sure you have `Last 12 months` selected
- Try to select last item in offered periods list (`Months this year`) - fails

### DIFF
--- a/packages/period-selector-dialog/package.json
+++ b/packages/period-selector-dialog/package.json
@@ -16,7 +16,7 @@
     "prebuild": "rimraf ./build/*",
     "build": "cross-env BABEL_ENV=es babel src --out-dir build --ignore spec.js && npm run build:css",
     "build:css": "cp -R ./css/* ./build/",
-    "watch": "npm run build --  --watch",
+    "watch": "npm run build --watch",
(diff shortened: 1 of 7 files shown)


packages/period-selector-dialog/package.json
```diff
@@ -16,7 +16,7 @@
     "prebuild": "rimraf ./build/*",
     "build": "cross-env BABEL_ENV=es babel src --out-dir build --ignore spec.js && npm run build:css",
     "build:css": "cp -R ./css/* ./build/",
-    "watch": "npm run build --  --watch",
+    "watch": "npm run build --watch",
```
amcgee: Note: I believe this will break with older verisons of `npm`, so we should probably leave it unless we're sure everyone's using a new-enough version.  What was the reason for removing the argument pass-through?

neeilya: Sorry, I was unaware of explicitly required `--`, thought it was copy paste error.
Will undo.

PR #350, opened 2018-11-22
Improve styling of Interpretations cards

amcgee:
In particular:

* Increase padding around `No Interpretations` text
* Fix card action hover, padding, margins, etc. and consolidate them into one `div`
* Use non-raised Mui `Card` style since we don't use raised cards anywhere else
* Refactor some DOM elements to use more idiomatic styles
* Improve padding around comment avatars

Also remove Mui deprecation warning from the example app

Before:
screenshot 2018-11-22 16 55 32
screenshot 2018-11-22 16 55 42
screenshot 2018-11-22 16 56 13

After:
screenshot 2018-11-22 16 51 56
screenshot 2018-11-22 16 52 03
screenshot 2018-11-22 16 52 31

### DIFF
--- a/examples/create-react-app/src/components/org-unit-dialog.js
+++ b/examples/create-react-app/src/components/org-unit-dialog.js
@@ -153,7 +153,7 @@ export default class OrgUnitDialogExample extends Component {
             <div style={{ padding: 16 }}>
                 <Button
                     onClick={this.toggleDialog}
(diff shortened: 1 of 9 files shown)


amcgee:
Note: There are quite a few things I would like to refactor / rewrite here but am leaving it largely intact for now to minimize the surface area of this aesthetic change

PR #352, opened 2018-11-23
fix/re-order build script so --watch applies to babel transpiling

jenniferarnesen:
Problem: js changes in period-selector-dialog and core were not re-transpiled when running `yarn watch`.  This is because the --watch option gets applied at the end of the command. When you have a compound npm script (command1 && command2), then the option is applied to the last command.

It is actually possible to insert the option using shell functions aka:
```"npm run lint -- -f unix"```
```"lint": "f() { eslint -f src $@ . && npm run tslint; }; f"```
credit: https://stackoverflow.com/questions/11580961/sending-command-line-arguments-to-npm-script?lq=1

But this seemed unnecessarily complex for this one case. 

Downside is that css changes dont trigger the build.

### DIFF
--- a/packages/period-selector-dialog/package.json
+++ b/packages/period-selector-dialog/package.json
@@ -13,8 +13,8 @@
     "react-dom": "^15.0.0 || ^16.0.0"
   },
   "scripts": {
-    "prebuild": "rimraf ./build/*",
-    "build": "cross-env BABEL_ENV=es babel src --out-dir build --ignore spec.js && npm run build:css",
+    "prebuild": "rimraf ./build/* && cross-env mkdirp ./build",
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/period-selector-dialog/package.json [hunk 1]
```diff
@@ -13,8 +13,8 @@
     "react-dom": "^15.0.0 || ^16.0.0"
   },
   "scripts": {
-    "prebuild": "rimraf ./build/*",
-    "build": "cross-env BABEL_ENV=es babel src --out-dir build --ignore spec.js && npm run build:css",
+    "prebuild": "rimraf ./build/* && cross-env mkdirp ./build",
```
amcgee: Is `cross-env` required with `mkdirp`?  I thought `mkdirp` was cross-platform

PR #355, opened 2018-11-26
remove minWidth and width rules, change to 'flexBasis', also set maxH…

joakimia:
Change rules to Paper components for MuiDialog:

Use FlexBasis instead of width such that Dialog sizes can be rendered dynamically and have its width flex freely based on available screen space.
Remove minWidth (will most likely cause problems on smaller screens).
Set maxHeight to 19% of total viewport (this value is something i just picked, after testing Dialogs on my 15 inch screen - not 100% sure if it is sufficient for smaller screen sizes) as the "AddToLayoutButton" on Dimension dialogs in DV crammed the dropDown Menu.
image

### DIFF
--- a/packages/core/src/theme/mui3.theme.js
+++ b/packages/core/src/theme/mui3.theme.js
@@ -96,19 +96,18 @@ export const theme = {
         },
         MuiDialog: {
             paperWidthSm: {
-                flexBasis: 400,
-                maxWidth: 400,
-                maxHeight: 'calc(100vh - 15%)',
+                flex: '0 1 400px',
+                width: '400px',
+                maxWidth: '400px',
             },
             paperWidthMd: {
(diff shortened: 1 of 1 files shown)


packages/core/src/theme/mui3.theme.js
```diff
@@ -96,19 +96,18 @@ export const theme = {
...
+                flex: '0 1 600px',
+                width: '600px',
+                maxWidth: '600px',
             },
             paperWidthLg: {
-                minWidth: 800,
-                width: 800,
-                maxWidth: 800,
+                flex: '0 1 800px',
+                maxWidth: '800px',
```
amcgee: @joakimia is `width` omitted from `paperWidthLg` on purpose?  Sorry about the delay in seeing this, been in meetings.

PR #360, opened 2018-11-28
feature/implement drag and drop to reorder selected periods in period-selector

jenniferarnesen:
Enables user to drag and drop one selected item at a time to reorder. Multi-drag will be implemented in a follow-up PR

### DIFF
--- a/packages/period-selector-dialog/src/PeriodsList.js
+++ b/packages/period-selector-dialog/src/PeriodsList.js
@@ -7,12 +7,12 @@ import {
 } from 'react-sortable-hoc';
 import PeriodListItem from './PeriodListItem';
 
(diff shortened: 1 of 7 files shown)


packages/period-selector-dialog/src/PeriodsList.js
```diff
@@ -1,33 +1,90 @@
...
     onRemovePeriodClick: PropTypes.func,
+    onReorder: PropTypes.func,
     listClassName: PropTypes.string.isRequired,
+    sortable: PropTypes.bool,
 };
 
 PeriodsList.defaultProps = {
     onPeriodDoubleClick: () => null,
     onRemovePeriodClick: () => null,
+    onReorder: () => null,
```
janhenrikoverland: Use Function.prototype for no-ops.

PR #369, opened 2018-12-06
chore/Use cra eslint configuration, fix easy lint warnings and errors, and add lint to build step

jenniferarnesen:
Fixes <JIRA ISSUE KEY>.

-
-
-

### DIFF
--- a/packages/header-bar/src/InnerHeader.js
+++ b/packages/header-bar/src/InnerHeader.js
@@ -2,7 +2,6 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import log from 'loglevel';
 import { Observable } from 'rxjs/Observable';
-import { Subject } from 'rxjs/Subject';
 import styles from './header-bar-styles';
 import getBaseUrlFromD2ApiUrl from './utils/getBaseUrlFromD2ApiUrl';
 
@@ -59,7 +58,7 @@ class InnerHeader extends Component {
 
     componentWillReceiveProps(props) {
         if (this.props.lastUpdate && (this.props.lastUpdate.getTime() - props.lastUpdate.getTime()) !== 0) {
-            dhis2.menu.ui.bootstrapMenu();
+            dhis2.menu.ui.bootstrapMenu();  //eslint-disable-line
         }
     }
 
--- a/packages/header-bar/src/menus/ProfileMenu.js
+++ b/packages/header-bar/src/menus/ProfileMenu.js
@@ -19,6 +19,7 @@ const ProfileMenu = addD2Context((props, { d2 }) => {
     }
 
     const initials = `${currentUser.firstName.charAt(0)}${currentUser.surname.charAt(0)}`;
+    /* eslint-disable */
     const rightSide = (
         <div style={styles.profileRightSide}>
             <div style={styles.profileFlexWrap}>
@@ -33,6 +34,7 @@ const ProfileMenu = addD2Context((props, { d2 }) => {
             </Button>
         </div>
     );
+    /* eslint-enable */
 
     return (
         <HeaderMenu
--- a/packages/table/src/__tests__/Table.component.spec.js
+++ b/packages/table/src/__tests__/Table.component.spec.js
@@ -1,10 +1,8 @@
 import React from 'react';
-import { describeWithDOM, shallow } from 'enzyme';
+import { shallow } from 'enzyme';
 import Table from '../Table.component';
 import TableHeader from '../TableHeader.component';
 import TableContextMenu from '../TableContextMenu.component';
-
-import Popover from 'material-ui/Popover/Popover';
 
 describe('Table component', () => {
     let tableComponent;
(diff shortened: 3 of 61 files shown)


jenniferarnesen:
Note that I have temporarily added a code in DeleteDialog.js that should cause a lint error and halt the build. Wanted to test that. Must remember to remove that code before merging!!

packages/table/src/__tests__/Table.component.spec.js
```diff
@@ -170,22 +168,22 @@ describe('Table component', () => {
 
         // TODO: The Popover requires a dom element as a targetEl prop. Figure out how to test this without a DOM.
         xit('should hide the contextmenu when left clicking outside the contextmenu', () => {
-            const fakeRowSource = { name: 'My item' };
+            // const fakeRowSource = { name: 'My item' };
```
amcgee: Why was this commented out?  Should it be fixed or removed instead?

packages/header-bar/src/menus/ProfileMenu.js
```diff
@@ -19,6 +19,7 @@ const ProfileMenu = addD2Context((props, { d2 }) => {
     }
 
     const initials = `${currentUser.firstName.charAt(0)}${currentUser.surname.charAt(0)}`;
+    /* eslint-disable */
```
amcgee: Why does eslint need to be disabled here?

packages/header-bar/src/InnerHeader.js
```diff
@@ -59,7 +58,7 @@ class InnerHeader extends Component {
 
     componentWillReceiveProps(props) {
         if (this.props.lastUpdate && (this.props.lastUpdate.getTime() - props.lastUpdate.getTime()) !== 0) {
-            dhis2.menu.ui.bootstrapMenu();
+            dhis2.menu.ui.bootstrapMenu();  //eslint-disable-line
```
amcgee: Why does eslint need to be disabled here?

jenniferarnesen: because `dhis2` is apparently a global. Actually, I will see if this code is triggered

amcgee:
Maybe better to define `dhis2` as an eslint global in this file rather than
disabling the whole line then?

packages/table/src/__tests__/Table.component.spec.js
jenniferarnesen: this block was commented out because `TableComponent` on line 175 is not recognized. The test is currently disabled, so I considered removing it alltogether, but decided to leave it as it seemed like a reasonable test case.

packages/header-bar/src/InnerHeader.js
varl: It's better to hint to eslint that `dhis2` is a global:

```
/* global dhis2 */
```

packages/header-bar/src/menus/ProfileMenu.js
varl: Also curious about which rule this violates?

On a related note: it is better to disable on a per-rule basis, that way other developers gets a hint about what rule it violates and could potentially fix it.

packages/header-bar/src/InnerHeader.js
jenniferarnesen: added /* global dhis2 */ and a check to be sure that it exists

packages/table/src/__tests__/Table.component.spec.js
jenniferarnesen: I have now removed this test, as it was not a good test (was trying to test functionality in one of the subcomponents). I also fixed some of the other tests, as they were not actually testing anything. Thanks for sending me down a rabbit hole!

packages/header-bar/src/menus/ProfileMenu.js
jenniferarnesen: Upon more investigation the code was pointlessly setting the `key` attribute with a non-existing variable "name". The `key` attribute has now been removed (can you just confirm that the attribute was pointless?)

packages/table/src/__tests__/Table.component.spec.js
amcgee: You're welcome :-)

packages/header-bar/src/menus/ProfileMenu.js
amcgee: Yes, looks pointless to me as well.  Maybe those used to be rendered as an array but as written now it's unnecessary to use a `key` attribute, especially one that's `undefined` 👍

packages/header-bar/src/menus/ProfileMenu.js
```diff
@@ -34,6 +35,7 @@ const ProfileMenu = addD2Context((props, { d2 }) => {
         </div>
     );
 
+
```
amcgee: Nitpick - unnecessary additional blank line?

packages/translation-dialog/src/TranslationForm.component.js
```diff
@@ -208,7 +208,7 @@ class WithObservableState extends Component {
         this.disposable = this.props.stateSource$
             .subscribe(
                 state => this.setState(state),
-                error => log.error(error),
+                error => console.log.error(error),
```
amcgee: Does `console.log.error` exist?  Isn't it just `console.error`?

packages/header-bar/src/menus/ProfileMenu.js
jenniferarnesen: Yes I know, it was pure laziness that I didn't specify the rule:  no-undef

jenniferarnesen: the "name" variable was not defined anywhere.

PR #372, opened 2018-12-19
Task/sort mentions list [DHIS2-5473]

joakimia:
This PR includes:
using ´lodash/sortBy` to sort available mentions by displayName.

image

Capital letters are prioritized, i checked the source code and it uses array.sort(), and MDN documentation states that:
```
The default sort order is built upon converting the elements into strings, then comparing their sequences of UTF-16 code units values.
```
This means that displayName with capital letters will be sorted first, and then displayNames without capital letters.

I confirmed this with concating the users with:
`
users.concat({id: 'testId', displayName: 'alain Traore', userCredentials: { username: 'traori' }})
`

I can change it to sort by `displayName.toLowerCase()`:

image

Additional info:
I've also updated the Typograhpy `variant` prop to use the new version as this was causing a warning.

* Test cases have been adjusted now that `UserList`  is rendering a sorted list.

* Added 1 test case to verify that the list is rendered in a sorted fashion.

### DIFF
--- a/packages/mentions-wrapper/src/UserList.js
+++ b/packages/mentions-wrapper/src/UserList.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import ListItemText from '@material-ui/core/ListItemText';
 import ListItem from '@material-ui/core/ListItem';
 import List from '@material-ui/core/List';
-import sortBy from 'lodash-es/sortBy';
+import sortBy from 'lodash/sortBy';
 
 import Popover from '@material-ui/core/Popover';
 import Typography from '@material-ui/core/Typography';
@@ -48,7 +48,7 @@ export const UserList = ({
 
         onSelect(user);
     };
-
+    
     const sortedUsers = sortBy(users, [userName => userName.displayName]);
 
     return (
--- a/packages/mentions-wrapper/src/UserList.js
+++ b/packages/mentions-wrapper/src/UserList.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import ListItemText from '@material-ui/core/ListItemText';
 import ListItem from '@material-ui/core/ListItem';
 import List from '@material-ui/core/List';
+import sortBy from 'lodash-es/sortBy';
 
 import Popover from '@material-ui/core/Popover';
 import Typography from '@material-ui/core/Typography';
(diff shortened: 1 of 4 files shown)


jenniferarnesen:
If I type in "@j", then I get this:

image


I'm thinking that John Kamara should come before John Traore.

packages/mentions-wrapper/src/UserList.js
```diff
@@ -47,6 +48,8 @@ export const UserList = ({
 
         onSelect(user);
     };
+    
+    const sortedUsers = sortBy(users, [userName => userName.displayName]);
```
jenniferarnesen: Consider moving this to line 75 rather than defining a var sortedUsers. There is no point calling sortBy if users is an empty array.

packages/mentions-wrapper/src/__tests__/UserList.spec.js
```diff
@@ -56,15 +57,25 @@ describe('Mentions: MentionsWrapper > UserList component', () => {
                 .find(ListItemText)
                 .first()
                 .props().primary
-        ).toEqual('Johnny Cash (sue)');
+        ).toEqual('Eric Clapton (slowhand)');
+    });
+
+    it('should sort the mentions in alphabetical order by firstname', () => {
+        const sortedUsers = sortBy(props.users, [userName => userName.displayName]);
```
jenniferarnesen: I don't think we should use the sortBy function in the test since that is the exact code in the implementation, thus not really testing anything. An alternative would be to spy on sortBy and ensure that it is called with the correct arguments. This is also a case where you could hardcode the expected result.

One other thing I want to mention regarding this test file is in setting the properties lines 24-37. I would define the users array separately, then in the props you will have:
```
{
users,
selectedUsers: users[2],
}
```

packages/mentions-wrapper/src/__tests__/UserList.spec.js
```diff
@@ -56,15 +57,25 @@ describe('Mentions: MentionsWrapper > UserList component', () => {
...
+            expect(
+                node.props().primary).toContain(sortedUsers[i].displayName
+                )
+            );
     });
 
     it('should trigger the onSelect callback when an item in the list is clicked', () => {
         const listItem = userList.find(ListItem).first();
-
+        const sortedUsers = sortBy(props.users, [userName => userName.displayName]);
```
jenniferarnesen: Same as above comment

PR #374, opened 2019-01-21
Small refactor

neeilya:
This PR contains small code refactoring in `packages/interpretations/src/models/Interpretation.js`

- Remove redundant async/await from delete/like methods
- Remove redundant .then(() => this)
- Return promise of the second request as a result in `save()` method

### DIFF
--- a/packages/interpretations/src/models/interpretation.js
+++ b/packages/interpretations/src/models/interpretation.js
@@ -33,7 +33,7 @@ export default class Interpretation {
             const sharingUrl = `/sharing?type=interpretation&id=${interpretationId}`;
 
             const sharingPayload = this.sharing
-                ? { object: { ...this.sharing, id: this.id } }
+                ? { object: { ...this.sharing, id: interpretationId } }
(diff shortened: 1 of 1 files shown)


packages/interpretations/src/models/interpretation.js
```diff
@@ -24,42 +25,40 @@ export default class Interpretation {
...
+
+            return apiFetch(d2, sharingUrl, "PUT", sharingPayload)
         } else {
-            return await apiFetch(d2, `/interpretations/${this.id}`, "PUT", this.text)
-                .then(() => {
-                    if (this.sharing) {
-                        const sharingPayload =  { object: {...this.sharing, id: this.id}};
-                        this.sharing = null;
+            // interpretation already exists in DB
+            await apiFetch(d2, `/interpretations/${this.id}`, "PUT", this.text);
```
jenniferarnesen: Is the purpose of this fetch to determine if the interpretation exists?

neeilya: I assume it is for updating the text of interpretation/comment.

PR #401, opened 2019-03-27
DHIS2-6207 - fix legend cancel bug

ismay:
fixes 6207 (https://jira.dhis2.org/browse/DHIS2-6207)

See the issue for more detail. Basically the assignment [here](https://github.com/dhis2/d2-ui/pull/401/files#diff-d8934815beb930fd28b94669c1debc61R80) was mutating state (not react state, but a custom store implementation). This didn't trigger a rerender, but would be visible when it rerendered for different reasons (like clicking the ellipsis to trigger the popup).

To fix this I shallow cloned the model so the assignment doesn't mutate the original. This however, caused a new bug because of the `Legend.component` `updateItem` method. This method compares the object references, so a clone is seen as a new item.

To fix that I changed `updateItem` to compare on the `id` prop, and not references. I'm being very conservative in my approach here, and not mutating the input array as I'm not sure if the input (which comes from [here](https://github.com/dhis2/maintenance-app/blob/master/src/config/field-overrides/legendSet.js#L25)) is safe to mutate. I also modified the naming slightly, as the way I read it the naming was off (`newItems` is actually `existingItems`).

### Test

Tested this by building d2-ui and copying `./lib` to the maintenance-app's `node_modules/d2-ui`, not ideal, but symlinking messed things up. On macos with the latest chrome and firefox the bug is resolved.

### DIFF
--- a/src/legend/Legend.component.js
+++ b/src/legend/Legend.component.js
@@ -78,18 +78,21 @@ class Legend extends Component {
     };
 
     updateItem = (existingItems) => {
-        let updatedItems = [];
-
+        let updatedItems = existingItems;
         const currentItem = legendItemStore.getState() && legendItemStore.getState().model;
-        const isNewItem = existingItems.every(item => item.id !== currentItem.id);
 
-        if (isNewItem) {
-            updatedItems = [...existingItems, currentItem];
-        } else {
-            updatedItems = existingItems.map((item) => {
-                const shouldBeUpdated = item.id === currentItem.id;
-                return shouldBeUpdated ? currentItem : item;
-            });
+        // Only update if we got a valid model from getState
+        if (currentItem) {
+            updatedItems = [...existingItems];
+            const idx = updatedItems.findIndex(item => item.id === currentItem.id);
+
+            if (idx === -1) {
+                // Add item if it's a new item
+                updatedItems.push(currentItem);
+            } else {
+                // Replace old item with changed item
+                updatedItems[idx] = currentItem;
+            }
         }
 
         return this.props.onItemsChange(updatedItems);
--- a/src/legend/Legend.component.js
+++ b/src/legend/Legend.component.js
@@ -77,14 +77,22 @@ class Legend extends Component {
         this.props.onItemsChange(newItems);
     };
 
-    updateItem = (newItems) => {
-        const modelToUpdate = legendItemStore.getState() && legendItemStore.getState().model;
-        const isNewLegendItem = newItems.every(model => model !== modelToUpdate);
+    updateItem = (existingItems) => {
+        let updatedItems = [];
 
-        return this.props.onItemsChange([].concat(
-            newItems,
-            isNewLegendItem ? modelToUpdate : [],
-        ));
+        const currentItem = legendItemStore.getState() && legendItemStore.getState().model;
+        const isNewItem = existingItems.every(item => item.id !== currentItem.id);
+
+        if (isNewItem) {
+            updatedItems = [...existingItems, currentItem];
+        } else {
+            updatedItems = existingItems.map((item) => {
+                const shouldBeUpdated = item.id === currentItem.id;
+                return shouldBeUpdated ? currentItem : item;
+            });
+        }
+
+        return this.props.onItemsChange(updatedItems);
     };
 
     // Check if end value is bigger than start value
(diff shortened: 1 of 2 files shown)


src/legend/Legend.component.js
```diff
@@ -77,14 +77,22 @@ class Legend extends Component {
...
-
-        return this.props.onItemsChange([].concat(
-            newItems,
-            isNewLegendItem ? modelToUpdate : [],
-        ));
+    updateItem = (existingItems) => {
+        let updatedItems = [];
+
+        const currentItem = legendItemStore.getState() && legendItemStore.getState().model;
+        const isNewItem = existingItems.every(item => item.id !== currentItem.id);
```
amcgee: I know this code was already existing, but this would be a little more efficient as `!existingItems.some(item => item.id === currentItem.id)` since it would short-circuit the iteration once a match was found.

src/legend/Legend.component.js
```diff
@@ -77,14 +77,22 @@ class Legend extends Component {
...
+
+        if (isNewItem) {
+            updatedItems = [...existingItems, currentItem];
+        } else {
+            updatedItems = existingItems.map((item) => {
+                const shouldBeUpdated = item.id === currentItem.id;
+                return shouldBeUpdated ? currentItem : item;
+            });
+        }
```
amcgee: Actually, you could rewrite this whole section a bit more clearly (and fewer loops through the array) as:

```
let updatedItems = existingItems;
const currentItem = legendItemStore.getState() && legendItemStore.getState().model

// This is necessary if the first half of the above expression ever evaluates to false,
// which it seems we expect it might?  Otherwise `currentItem.id` below will throw an error.
if (currentItem) {
  updatedItems = [...existingItems];
  const idx = updatedItems.findIndex(item => item.id === currentItem.id)
  if (idx === -1) {
    updatedItems.push(currentItem)
  } else {
    updatedItems[idx] = currentItem
  }
}

return this.props.onItemsChange(updatedItems);
```

Not a required change, just thoughts while thinking

src/legend/LegendItem.store.js
```diff
@@ -72,10 +72,10 @@ const formFieldsConfigs = [{
     }), ColorPicker),
 }];
 
-
 // Called when a field is changed
 export function onFieldChange(fieldName, value) {
-    const model = legendItemStore.getState().model;
+    // Shallow clone model, so the assignment below doesn't mutate the original
+    const model = { ...legendItemStore.getState().model };
```
HendrikThePendric: If this is in fact a d2 Model instance, this could be problematic. This shallow clone will be a POJO containing only the instance's properties, but not the class methods.

In this past I have encountered a similar problem and solved it differently. I believe like this:
```javascript
const originalModel = legendItemStore.getState().model;
// This creates a fresh model instance with all the properties of the old model 
// and all of the class methods are still there
const model = originalModel.modelDefinition.create(originalModel..toJSON());
model[fieldName] = value
```

ismay: > If this is in fact a d2 Model instance

I believe it is actually, hmm.

ismay: This code is so full of footguns. I'll check the prototype of the model to verify, thanks for the heads up.

src/legend/Legend.component.js
ismay: Nice, though I'm not sure about `findIndex` as we might eventually decide to use only current syntax. Your addition of a failsafe for a missing model is a good thing to do anyway, I'll add that, thanks!

src/legend/LegendItem.store.js
ismay: I've checked, but it doesn't seem to be an instance of a d2 Model:

```
Object.getPrototypeOf(originalModel)
> Object { … }
```

So that's good 🙂

src/legend/Legend.component.js
ismay: Though on the other hand currently we're using stage-0 with babel, so maybe that's a bit too careful.

src/legend/LegendItem.store.js
HendrikThePendric: That's good 👍

src/legend/Legend.component.js
amcgee: findIndex is in the ES2015 standard so I think it's fine to use.  We need to polyfill things for IE anyway, just worth making sure we're doing that :-)

ismay: You're right, I've added your suggested changes. Thanks!

PR #416, opened 2019-04-10
fix: cancel create legends

ismay:
Fixes: https://jira.dhis2.org/browse/DHIS2-6353

The cancel button didn't cancel. Which makes sense, because it was exactly the same logic as the proceed button.

### DIFF
--- a/src/legend/Legend.component.js
+++ b/src/legend/Legend.component.js
@@ -130,7 +130,11 @@ class Legend extends Component {
         this.setState({ warningDialogOpen: true });
     };
 
-    handleClose = () => {
+    handleCreateLegendsCancel = () => {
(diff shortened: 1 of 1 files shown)


ismay:
@Mohammer5 It's basically just this dialog, so the user doesn't have to enter any values:

Screenshot 2019-04-10 at 11 46 41

Not sure about the onRequestClose prop though, I assume it's meant for clicks outside of the dialog.

Mohammer5:
@ismay You're right! I was talking nonsense :)
I second your assessment of what `onRequestClose` means! Go for it

ismay:
😊 Cool, thanks!

ismay:
To be merged in #417, #418 and #419 after the freeze.

PR #424, opened 2019-04-25
fix: set menu prop disableAutoFocusItem to true by default

janhenrikoverland:
Fixes https://jira.dhis2.org/browse/DHIS2-6616 and https://jira.dhis2.org/browse/DHIS2-6654

- Set Menu prop "disableAutoFocusItem" to true by default, ref https://material-ui.com/api/menu


amcgee:
I think the overrides here only work for css, not component properties

PR #428, opened 2019-04-29
Feat/period picker as a package

HendrikThePendric:
### UPDATE
Based on @jenniferarnesen comment about aligning the period-picker and the period-selector, I have come up with a few ideas to make the way the PeriodPicker works more similar to how the period-selector works. I'd like to implement them first, and take it from there. So for now, I think we can put this PR "on-hold".

### Changes proposed in this pull request:

- Introduces the PeriodPicker component as a package, as opposed to a part of ui-core. This seems like a sensible thing to do, since it's a fairly complex component that contains a lot of business logic.
- The component now includes a field for `periodType`, as well as the fields to create an actual period (periodId). This makes sense because:
    - The `periodType` value is never used by itself. The only value relevant to the server is the `periodId` (`props.value`)
   - The available `periodTypes` can be fetched from the server, but in order to do anything with these types, support needs to be added to the `PeriodPicker` component. So by integrating the `periodType` drop-down into the component and having the component fetch the available `periodTypes` we can take back control of this.

### Features of the new component:
- Fetches available period-types from the server
- Throws a warning in development environment if period-types are returned from the server that are not supported by the component
- Is responsible for its own translations, using `@dhis2/i18n`
- Leverages the d2 `period-parser` to validate `periodId` values and produce human readable period-descriptions
- Includes validation to catch invalid day numbers for a month, or invalid week/bi-week numbers for a year.
- Pretty decent unit-test coverage
- **And my personal favourite:** the component behaves like a regular controlled-component form-field, so it has a `value` and an `onChange` prop (to update the value) . When the `value` changes, the related state values are updated automatically (i.e. if you set the value to `2016ThuW4` the state value for periodType will switch to WeeklyThursday, and the state values for year and week are set to 2016 and 4 respectively). I think this will make it much easier for this input to work together with any given form-library.

### Know issues
I've created this PR as a draft, because there are still two things wrong:
1. I could not get the `examples/create-react-app` to run with all the other components in there, so while I was working I just commented out all the other stuff. I'm not sure if this should be looked into in the current PR. I hope not.
1. ~[The build is failing](https://travis-ci.com/dhis2/d2-ui/builds/109980061). I get an error that I can't reproduce locally. For me all tests are passing. Not sure how to deal with this either...~ FIXED now

Looking forward to your input.


jenniferarnesen:
@HendrikThePendric Apologies if I ask the obvious, but is this code that was moved out of d2-ui-core and into its own package?

jenniferarnesen:
I get that we want to pull good components out of d2-ui-core, but I'm wary of a proliferation of packages.

In this notes issue: https://github.com/dhis2/notes/issues/20 there has been a discussion of what to do with d2-ui. One of my last comments suggests a path towards a single package with these complex components that are d2 dependent (and therefore do not go into `ui-widgets`).

jenniferarnesen:
Oh, and I'm also wondering if we should somehow align the period-picker and the period-selector components. Seems that both are generating period names, and I'm guessing they have some common functionalities.

HendrikThePendric:
> , but is this code that was moved out of d2-ui-core and into its own package?

Yes and no. I've left the PeriodPicker in `d2-ui-core` as it was (for now). This PR is a newer version, which only uses some logic of the original component. The main reason why I didn't replace the "old" period-picker is because this new one incorporates a dropdown for period-type, while the old one just displays the inputs to generate a periodId. The newer version also relies on having d2 passed as a prop, and my understanding was that anything that relies on d2 should go into packages.

> suggests a path towards a single package with these complex components that are d2 dependent (and therefore do not go into `ui-widgets`)

Happy to move this to wherever. I thought that since this component relies on MUI v3 and takes d2 as a prop, this would be suitable place. Especially because a lot of other packages in d2-ui are the same in that respect (i.e. file-menu, sharing-dialog, etc). Where would you suggest we should put this?

> we should somehow align the period-picker and the period-selector components

Yes, good suggestion. I guess I should have discussed this before I got started. Both are distinct components, but they could share some underlying logic/features:
- The PeriodSelector's year dropdown with the two arrows to navigate through sets of years is very good and should really be used in both components.
- I'm not sure about this yet, but perhaps the PeriodPicker should also show the generated periods as the PeriodSelector does. So this would mean they could share the FixedPeriodsGenerator. However, I don't think this has been implemented in the best possible way. There is a period-parser in d2 which could have done most of the hard work.
- One of the reasons I started with the PeriodPicker was because I kept finding that new period-types had been added on the server and the component doesn't support this type yet. So now I let the component fetch the periodTypes from the server and throw a developer-warning if unsupported periods are found. I think it would be good if this mechanism was shared by the PeriodSelector (somehow?)... And if we could share that FixedPeriodsGenerator (or whatever we want to call it) between components, than we could at least always make sure that both components support the same period-types....

HendrikThePendric:
Hi @varl & @jenniferarnesen
I've had a go at aligning the period picker with the period-selector. I've made quite a lot of changes, but I think the bottom line is this:
- I now have a mechanism for generating period fields that is capable of powering both the period-picker and the period-selector. I am now using generated periods in the period-picker, same as in the period selector. My implementation supports all currently available period-types, and leverages the d2 period-parser to generate the IDs. By using the period parser, I was able to generate periods with much less code. This will come in handy when we have to add new periodTypes.
- I have updated the year select in the period-picker so that it now also has "shift forth" and "shift back" buttons to update the date ranges, same as the period-selector does....

So, now the question is how to move forward. To me it would make sense if these components are all the same package. The current package name would not be very descriptive anymore... And obviously, some work would have to be done to actually integrate these two packages.

Alternatively, we could perhaps argue that we could keep the components in separate packages. And that the "period field generator" is moved to d2 (?). Less keen on this option, because when creating options, we are also creating localised strings (`@dhis2/i18n`), and I don't think that should happen in d2...

Looking forward to finding out your opinions

varl:
>     * 
> 
> 
> So, now the question is how to move forward. To me it would make sense if these components are all the same package. The current package name would not be very descriptive anymore... And obviously, some work would have to be done to actually integrate these two packages.

Good job on this. I don't have a great name on hand for this, but usage-wise having them in the same package makes sense.

```js
import { PeriodSelector, PeriodPicker } from '@dhis2/d2-ui-period-stuff`
```


... What's the difference between a **Picker** and **Selector** for **Periods**? No idea. As I said, I don't have great names on hand for these, but to me picker and selector represent the same thing so we need to figure something out. :D

HendrikThePendric:
Maybe it'l help if I outline the differences, to get to good descriptive naming:
- Period-selector is multi-select, period-picker is single-select.
- Period-selector does relative and fixed periods, period-picker only does fixed periods.
- Because the period-selector lets users select/deselect multiple periods, and you can even mix fixed and relative periods, it uses the `ItemSelector` component, which makes it take up a lot more screen real-estate. I think that is why this component is also exporting the PeriodSelectorDialog because it might be too big in some cases. The PeriodPicker is just a bunch of Selects (max 4 selects on 2 rows) so takes up a lot less space. It's much more of an "inline form component".

So perhaps the component names are descriptive enough. A selector (could) imply you select multiple items. However when you "pick" something it typically is one item. The package name should be made a little bit more generic though. Since we also have a package called "rich-text" and "legend", why don't we just call this package "period":
```javascript
import { PeriodSelector, PeriodPicker } from '@dhis2/d2-ui-period`
```

varl:
Ok, thanks. Yeah, package name needs to be a bit more generic. I don't see any deletions from d2-ui-core so this could probably slip in there with a minor version bump?

HendrikThePendric:
Update: @jenniferarnesen and I will discuss the way to consolidate these two packages next week (12/6) when I'm in Oslo. Putting this on hold for now.

HendrikThePendric:
This PR can be closed now. We have had meetings about the possibles solution and this problem needs to be addressed more thoroughly. We need:
- Multi-calendar support in the client
- Various date/period related components by this multi-calendar engine

See https://github.com/dhis2/notes/issues/51

This PR does not meet these criteria

PR #429, opened 2019-05-02
fix(richtext): skip parsing text in links

edoardo:
Fixes DHIS2-6821.

If links are detected in the text, skip parsing emphasis and emoji in
them.
A list of link is generated and during parsing the current position of
the marker is checked against the start and end position of links.
If the marker's position falls in any of the links' range, skip parsing.

### DIFF
--- a/packages/rich-text/src/parser/MdParser.js
+++ b/packages/rich-text/src/parser/MdParser.js
@@ -36,10 +36,22 @@ const codes = {
     },
 };
 
+let md;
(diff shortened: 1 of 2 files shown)


edoardo:
We found some edge cases that don't work well.
For example example_tricky_1.com/path and in general URLs with underscore in the hostname.

edoardo:
Fixed the edge cases of the comment above.

PR #431, opened 2019-05-16
chore: remove ControlBar component from d2-ui-core

jenniferarnesen:
BREAKING CHANGE: the ControlBar component has been removed from `d2-ui-core`.

This component was built specifically for Dashboards app, and no other DHIS2 Core app uses it. The component was therefore moved to the Dashboards app.

This change will cause d2-ui packages to be bumped from 5.3.x to 6.0.0


varl:
:tada:

PR #445, opened 2019-08-13
chore: transifex configuration

Philip-Larsen-Donnelly:


Philip-Larsen-Donnelly:
There aren't any localisations in v29 so this PR may be unnecessary.

PR #446, opened 2019-10-02
fix: support multiple ou roots in OrgUnitSelector component

edoardo:
Fixes DHIS2-7679.

- allow propagation of multiple ou roots from apps to `OrgUnitTreeMultipleRoots` used internally by `OrgUnitSelector`

### DIFF
--- a/packages/org-unit-dialog/src/OrgUnitSelector.js
+++ b/packages/org-unit-dialog/src/OrgUnitSelector.js
@@ -227,6 +227,9 @@ class OrgUnitSelector extends Component {
                                 }
                                 showFolderIcon
                                 disableSpacer
+                                useUserDataViewFallback={
+                                    this.props.useUserDataViewFallback
+                                }
                             />
                             <Menu
                                 anchorEl={this.state.menuAnchorElement}
@@ -359,6 +362,8 @@ OrgUnitSelector.propTypes = {
      * Checkbox color in org unit tree
      */
     checkboxColor: PropTypes.string,
+
+    useUserDataViewFallback: PropTypes.bool,
 };
 
 OrgUnitSelector.defaultProps = {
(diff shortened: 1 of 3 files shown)


packages/org-unit-dialog/src/OrgUnitSelector.js
```diff
@@ -173,26 +192,39 @@ class OrgUnitSelector extends Component {
...
+                                    styles.orgUnitTree.selectedLabelStyle
+                                }
                                 checkboxColor={this.props.checkboxColor}
-                                displayNameProperty={this.props.displayNameProperty}
+                                displayNameProperty={
+                                    this.props.displayNameProperty
+                                }
                                 showFolderIcon
                                 disableSpacer
                             />
```
varl: This is the only functional change from what I can see. Is that correct?

edoardo: Sorry, after the last commit I lost reference here.
I believe you are referring to the addition of the {{roots}} prop, yes, that is the change.
I kept the prettier commit separated, so it's easier to see the actual change in commit: https://github.com/dhis2/d2-ui/pull/446/commits/936c2ba2bb9eff65824a247bf37224236f7d9bfa

packages/org-unit-tree/src/utils.js
```diff
@@ -62,7 +82,12 @@ export const mergeChildren = (root, children) => {
  * @param forceReloadChildren
  * @returns {*}
  */
-export const loadChildren = (root, displayNameProperty, forceReloadChildren) => {
+export const loadChildren = (
+    root,
+    displayNameProperty,
+    forceReloadChildren,
+    useUserDataViewFallback
```
janhenrikoverland: Maybe call it "isUserDataViewFallback" so it doesn't look like a custom react hook.

edoardo: ok

edoardo: Changed

PR #490, opened 2019-11-04
fix: avoid wrapping period type label on multiple lines

edoardo:
Fixes DHIS2-7609.

- avoid wrapping the period type label on multiple lines
- cleanup unused styles


edoardo:
> LGTM
> 
> One question- is the i18n-generate upgrade unrelated?

Yes, but needed for the build due to new translations added to the interpretations package.

PR #521, opened 2019-12-13
fix: update version to 6.3.3

edoardo:
Fixes DHIS2-7728.

- version bump


packages/file-menu/src/DeleteDialog.js
```diff
@@ -64,7 +64,14 @@ DeleteDialog.defaultProps = {
 
 DeleteDialog.propTypes = {
     open: PropTypes.bool,
-    fileType: PropTypes.oneOf(['chart', 'eventChart', 'reportTable', 'eventReport', 'map']),
+    fileType: PropTypes.oneOf([
```
martinkrulltott: can we save all of these as a global const somewhere to avoid repetition?

edoardo: I added the list in `util.js`.

PR #540, opened 2020-01-17
feat: favorites dialog filter by type

edoardo:
Fixes DHIS2-8156.

- allow to filter the list by visualization type


packages/favorites-dialog/src/icons/YearOverYearLineIcon.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import SvgIcon from '@material-ui/core/SvgIcon';
 
-export default <SvgIcon viewBox="0 0 16 16">
+export default <SvgIcon viewBox="0 0 16 16" style={{ marginRight: 8 }}>
```
martinkrulltott: why hardcode a margin on the icon itself instead of putting it on the wrapping container / where it's being used?

edoardo: I played around with various solutions with MUI components but ended up with the simplest.
I now added a wrapper element around the icons and apply the margin there.

martinkrulltott: Awesome! 🚀

PR #543, opened 2020-01-28
fix: fix interpretation comment delete

edoardo:
Fixes DHIS2-8118.

- fix the interpretation comment delete

Before it was always deleting the last added comment, regardless of
which delete button was clicked.
Fixed by moving the DeleteDialog in the comments list component, we only
need one dialog anyway.
The comment to delete is set in the state via callback when the Delete button
on a comment is used.
When the dialog is confirmed the comment to delete is retrieved from the
state.

### DIFF
--- a/packages/interpretations/src/components/Comment/__tests__/Comment.spec.js
+++ b/packages/interpretations/src/components/Comment/__tests__/Comment.spec.js
@@ -2,7 +2,6 @@ import React from 'react';
 import { shallow } from 'enzyme';
 import { Comment } from '../Comment';
 import ActionButton from '../../Buttons/ActionButton';
-import DeleteDialog from '../../DeleteDialog/DeleteDialog';
 
 let shallowComment;
 
@@ -21,9 +20,6 @@ const baseProps = {
     onEdit: jest.fn(),
     onReply: jest.fn(),
     onDelete: jest.fn(),
-    dialogIsOpen: false,
-    onDeleteConfirm: jest.fn(),
-    onDeleteCancel: jest.fn(),
 };
 
 const comment = (partialProps = {}) => {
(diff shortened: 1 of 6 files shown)


packages/interpretations/src/components/Lists/CommentsList.js
```diff
@@ -58,7 +60,9 @@ export class CommentsList extends React.Component {
         comment.save(this.context.d2).then(() => this.props.onChange(this.props.interpretation));
     };
 
-    onDeleteComment(comment) {
+    onDeleteComment() {
+        const comment = this.state.commentToDelete;
+        console.log('comment to delete', comment);
```
martinkrulltott: I guess this was left over from testing?

martinkrulltott: If so, the `const` isn't needed since it's only used once.

packages/interpretations/src/components/Lists/CommentsList.js
```diff
@@ -154,9 +155,17 @@ export class CommentsList extends React.Component {
...
-                {Comments}
-                {InputField}
+            {ViewMoreReplies}
+            {Comments}
+            {InputField}
+            {this.state.deleteDialogIsOpen && (
+                <DeleteDialog
+                    title={i18n.t('Delete comment')}
+                    text={i18n.t('Are you sure you want to delete this comment?')}
+                    onDelete={() => this.onDeleteComment(this.state.comment)}
```
martinkrulltott: `this.state.comment` is being passed to `onDeleteComment `  but it looks like`onDeleteComment` doesn't take any input params? (row 63: `onDeleteComment() {`)

packages/interpretations/src/components/Comment/__tests__/Comment.spec.js
```diff
@@ -112,20 +108,4 @@ describe('components: Comment -> Comment component ', () => {
         });
     });
 
-    describe('with prop dialogIsOpen as false', () => {
```
martinkrulltott: Can these deleted tests be replaced with something else or are they completely obsolete?

martinkrulltott: Or is that what's now being tested in `CommentsList.spec.js`?

edoardo: It's being tested in `CommentsList.spec.js`.

packages/interpretations/src/components/Lists/CommentsList.js
edoardo: yep, i'm going to clean it up.

packages/interpretations/src/components/Lists/CommentsList.js
edoardo: Right, another thing left over. Cleaning up.

PR #557, opened 2020-02-06
fix: file menu and open dialog fixes

edoardo:
Fixes https://jira.dhis2.org/browse/DHIS2-8250.

- use smaller size for vis type icons
- remove "Open in API" link in File > Get link


edoardo:
Yes, I thought about it, we have something similar in DV.
But all this is temporary and is going to be refactored/replaced soon.
The whole component is going to be rewritten from scratch as soon as the required ui-core pieces are in place.
The icons are now duplicated in different places/apps (DV, dashboards, analytics), we should have them in 1 place and import from there (see also my comment here: https://github.com/dhis2/dashboards-app/pull/529#discussion_r376252010

PR #567, opened 2020-02-17
fix: upgrade dhis2/analytics in period-selector-dialog

jenniferarnesen:
Upgrade to the latest @dhis/analytics library. No new features or bug fixes. This is an attempt to resolve an issue in dashboards where a seemingly wrong version of ui-core is causing a showstopping error. It seems really unlikely that the version of ui-core in period-selector (via @dhis2/analytics) would be the cause, but keeping dependencies current is usually a good idea anyway, right? :)

### DIFF
--- a/packages/period-selector-dialog/package.json
+++ b/packages/period-selector-dialog/package.json
@@ -1,47 +1,47 @@
 {
-  "name": "@dhis2/d2-ui-period-selector-dialog",
-  "description": "Period selector dialog component for DHIS2",
-  "main": "./build/index.js",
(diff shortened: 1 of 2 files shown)


jenniferarnesen:
Upgrade to the latest @dhis/analytics library. No new features or bug fixes. This is an attempt to resolve an issue in dashboards where a seemingly wrong version of ui-core is causing a showstopping error. It seems really unlikely that the version of ui-core in period-selector (via @dhis2/analytics) would be the cause, but keeping dependencies current is usually a good idea anyway, right? :)

PR #572, opened 2020-02-28
fix(data table): transform indonesian locale underscore to dash

Mohammer5:
Relates to DHIS2-7698.

This change will handle `in_ID` locales correctly and transform them to `in-ID` in the `DataTable` component.

It's the only component using the `keyUiLocale` value from the user settings, if this is the cause for all the issues in DHIS-7698, this could potentially solve a lot of issues.

### DIFF
--- a/src/data-table/data-value/valueRenderers.js
+++ b/src/data-table/data-value/valueRenderers.js
@@ -48,7 +48,22 @@ class DateValue extends PureComponent {
         // Get the locale from the userSettings
         this.context.d2.currentUser.userSettings
             .get('keyUiLocale')
-            .then(uiLocale => this.setState({ uiLocale }));
+            .then(uiLocale => {
+                /**
+                 * Underscore locales are causing the Intl library
+                 * to throw an error which breaks apps.
+                 * This is just a hack for the Indonesian locale.
+                 *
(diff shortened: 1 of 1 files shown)


src/data-table/data-value/valueRenderers.js
```diff
@@ -48,7 +48,13 @@ class DateValue extends PureComponent {
         // Get the locale from the userSettings
         this.context.d2.currentUser.userSettings
             .get('keyUiLocale')
-            .then(uiLocale => this.setState({ uiLocale }));
+            .then(uiLocale => {
+                if (uiLocale === 'in_ID') {
```
varl: Please add a comment with a reference to the JIRA ticket so it's possible to follow a breadcrumb trail to the root issue, with a brief explanation of why the hack is here and on which criteria we can remove it.

Mohammer5: Done. Please have another look @varl

varl: Perfect, thanks!

varl:
@Mohammer5 LGTM

There is instructions on how to cut a release here: https://github.com/dhis2/d2-ui/blob/master/docs/RELEASING.md

PR #574, opened 2020-03-06
fix: reorganize vis types in FileOpen dialog filter (DHIS2-8413)

edoardo:
Fixes https://jira.dhis2.org/browse/DHIS2-8413.

- implement the new design for the filter by visualization type in the File Open dialog.

"Chart" and "Pivot Table" are at the top of the list.
"Chart" comprises of all chart types, basically all the types listed after "Pivot Table".
There are now 4 filtering options, all visualizations, all chart visualizations, only pivot table visualizations, only a specific chart type visualizations.
The order of the chart types is also the same as the one in the visualization type selector in DV.

Before:
Screenshot 2020-03-06 at 10 50 23

After:
Screenshot 2020-03-06 at 11 15 06


packages/favorites-dialog/src/EnhancedToolbar.js
```diff
@@ -31,6 +32,13 @@ const toolbarStyles = () => ({
     },
 });
 
+const VisTypeFilterMenuItem = withStyles(toolbarStyles)(({ classes, type, icon = undefined, label = undefined }) => (
```
martinkrulltott: Whats the difference between `x = undefined` and just leaving it as `x`? Wouldn't it fallback to `undefined` if not provided?

packages/favorites-dialog/src/actions/index.js
```diff
@@ -126,11 +128,20 @@ export const fetchData = () => {
...
+            switch (state.filtering.visTypeValue) {
+                case 'all':
+                    break;
+                case CHART:
+                    favoriteModel = favoriteModel
+                        .filter()
+                        .on('type')
+                        .notEqual(PIVOT_TABLE);
+                    break
+                default:
```
martinkrulltott: Maybe not the prettiest logic to use whatever was provided to the switch case `(default)` as a filter as the fallback, but I guess only valid input will be passed in since it's based on the dropdown, not user input.

martinkrulltott: But if you want to be safe you could check if the input is present in `visTypeMap` though...

PR #591, opened 2020-04-16
fix: set d2-i18n to peer dependency

jenniferarnesen:
BREAKING CHANGE: apps that use components having i18n must have d2-i18n as a dependency

Part of fix for https://jira.dhis2.org/browse/DHIS2-8638:

* Set d2-i18n to peer dependency for components that have their own translations
* Use d2-app-scripts to extract and generate the translation files. This is to remove the error-prone setting of the default namespace and language.
* upgrade to node 10, required by d2-app-scripts, and we want to be current
* upgraded node-sass to the version that is compatible with node 10
* add d2-i18n as dependency in the example app

Note: my editor reformatted all the package.json files so the diff should be viewed with `w=1` query param in the url


packages/interpretations/src/components/InterpretationsComponent/InterpretationsComponent.js
```diff
@@ -8,14 +8,8 @@ import pick from 'lodash/fp/pick';
 import { getFavoriteWithInterpretations } from '../../api/helpers';
 import Details from '../DetailsPanel/Details';
 import InterpretationsCard from '../Cards/InterpretationsCard';
-import i18n from '../../locales';
```
amcgee, line 11: Unfortunately I think this import needs to still be somewhere - it's a module with sideeffects (it adds the .json resources on load, should be added to the `package.json` [`sideEffects` array](https://webpack.js.org/guides/tree-shaking/).  I think if we remove this import the locale strings for this package will never get loaded?  Unless of course it's imported somewhere else (possible, I didn't search) in which case this is fine!

amcgee: See [the locales/index.js template](https://github.com/dhis2/app-platform/blob/master/cli/src/lib/i18n/templates/locales.hbs#L15-L18)

amcgee:
Hey @jenniferarnesen - I think we need to add `sideEffects` arrays to all the `package.json` files too, otherwise the `import './locales.json'` could get eliminated by webpack?  Strange that it worked with the CRA example, though...

jenniferarnesen:
> Hey @jenniferarnesen - I think we need to add `sideEffects` arrays to all the `package.json` files too, otherwise the `import './locales.json'` could get eliminated by webpack? Strange if it worked with the CRA example, though...

Good idea. The packages seemed to work (see below), but it is more robust to add it as a side effect.

I didn't actually try it in the CRA (doesn't look like that app gets the userSettings). I built and published an alpha for each of the packages and used in my local DV. Here is the FileMenu in Spanish:  
image

And the interpretations menu:
image

amcgee:
Are the strings `No description` and `Read/Write` in the above screenshot untranslated for some reason?

jenniferarnesen:
> Are the strings `No description` and `Read/Write` in the above screenshot untranslated for some reason?

Good questions. I looked up Read/Write, which is in this file: https://github.com/dhis2/d2-ui/blob/master/packages/interpretations/src/sharing/sharingText.js#L3

What I think is that the const is evaluated before the language switch, and is never evaluated again since it is not in a function that is run at runtime. I've seen several cases of this, and we should definitely go through our libs and apps and make sure to fix it. Or maybe with a new i18n architeture this wouldn't be necessary, not sure.

amcgee:
OK.  We should look into fixing those.  Maybe the `_` is the problem? The rendered string should probably be wrapped in an `<em>` tag instead.

See this related Maps PR comment from today : https://github.com/dhis2/maps-app/pull/590#pullrequestreview-398052366

amcgee:
If needed we can fix those in a separate PR, but if it's easy we should do them here.

jenniferarnesen:
Ha I just figured out now that the underscores were for emphasis. Yes, definitely better to use an html tag 💯

jenniferarnesen:
I'll add a tech debt jira issue for these things. (maybe subtasks for each of the components/apps)

PR #621, opened 2020-08-14
feat: support Stacked area vis type in open dialog (DHIS2-9217)

edoardo:
Fixes [DHIS2-9217](https://jira.dhis2.org/browse/DHIS2-9217).

- add support for filtering Stacked area visualizations in the Open dialog (favorites-dialog package)

### DIFF
--- a/packages/favorites-dialog/i18n/en.pot
+++ b/packages/favorites-dialog/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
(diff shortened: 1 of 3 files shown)


martinkrulltott (approved):
Change LGTM. This has probably been discussed before, but the consts in in `favorites-dialog/src/visTypes.js` can't be fetched from Analytics instead? [example](https://github.com/dhis2/analytics/blob/master/src/modules/visTypes.js#L33). Probably out of scope for this PR anyway, just a thought.

edoardo:
> Change LGTM. This has probably been discussed before, but the consts in in `favorites-dialog/src/visTypes.js` can't be fetched from Analytics instead? [example](https://github.com/dhis2/analytics/blob/master/src/modules/visTypes.js#L33). Probably out of scope for this PR anyway, just a thought.

Yes, can definitely be done.
But since this component should be rewritten from scratch as soon as we have some time, I went with the quick fix, without adding another dependency.

PR #624, opened 2020-09-03
fix: correctly detect italic/bold marker after non-word char  (DHIS2-9382)

edoardo:
Fixes [DHIS2-9382](https://jira.dhis2.org/browse/DHIS2-9382).

- fix regexp for correctly detecting the closing italic/bold marker when is preceded by a non-word character.

Fixes the specific case reported in the ticket and any other string where a non-word character is the last character before the closing italic/bold marker.


janhenrikoverland (approved):
LGTM

PR #626, opened 2020-09-08
fix(interpretations): compute avatar initials from displayName

edoardo:
`firstName` and `surname` are not returned from the API anymore.
Compute the initials for the avatar from `displayName` instead.

### DIFF
--- a/packages/interpretations/src/components/Avatar/UserAvatar.js
+++ b/packages/interpretations/src/components/Avatar/UserAvatar.js
@@ -4,16 +4,21 @@ import Avatar from '@material-ui/core/Avatar';
 import { withStyles } from '@material-ui/core/styles';
 import styles from './styles/Avatar.style';
 
-const UserAvatar = ({ classes, firstName, surname }) => {
-    const initials = firstName.charAt(0).concat(surname.charAt(0))
+export const UserAvatar = ({ classes, displayName }) => {
+    const nameParts = displayName ? displayName.split(' ') : ['USER']
+
+    let initials = nameParts.shift().charAt(0)
+
+    if (nameParts.length) {
+        initials += nameParts.pop().charAt(0)
+    }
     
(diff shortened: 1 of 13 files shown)


packages/interpretations/src/components/Avatar/UserAvatar.js
```diff
@@ -4,16 +4,21 @@ import Avatar from '@material-ui/core/Avatar';
 import { withStyles } from '@material-ui/core/styles';
 import styles from './styles/Avatar.style';
 
-const UserAvatar = ({ classes, firstName, surname }) => {
-    const initials = firstName.charAt(0).concat(surname.charAt(0))
+export const UserAvatar = ({ classes, displayName }) => {
+    const nameParts = displayName ? displayName.split(' ') : ['USER']
```
janhenrikoverland, line 8: This will display "U" in all languages if there's no displayName. Thoughts @cooper-joe ?

edoardo: I used the same approach as in the interpretation app. See: https://github.com/dhis2/interpretation-app/commit/1a50e031552022407110c736b4589c40b68c3236#diff-9e2af752db61f9b9e0b80c4e84abe3c5R115

cooper-joe: Ideally this would use the Design System [avatar component](https://github.com/dhis2/design-system/blob/master/atoms/avatar.md) which falls back to an icon if there is no valid initials. However, that component is not yet built.

I think displaying "U" for user is a decent solution - there isn't any better letter/symbol I can think of. We could use the unicode "☺︎" or just a blank symbol like "☐", but I'm not certain how they'll display across languages. So it's probably safer to use what we've been using in interpretations until we use a proper component.

janhenrikoverland (approved):

PR #628, opened 2020-09-14
fix: org-unit-dialog z-index issue

martinkrulltott:
Bumps the z-index of the right-click menu and the two select elements in the org-unit-dialog to be compatible with @dhis2/ui z-index ranges. When used within a @dhis2/ui `Modal` (z-index 3000) the z-index needs to be higher than the modal's, which is why 3100 is used.

-----------

image
image
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dhis2/d2-ui",
-  "version": "7.0.7",
+  "version": "7.0.8",
   "license": "BSD-3-Clause",
   "private": true,
   "scripts": {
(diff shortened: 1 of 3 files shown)


edoardo (approved):
LGTM.

janhenrikoverland (approved):

martinkrulltott:
Note: when adding this to DV, this should be removed: https://github.com/dhis2/data-visualizer-app/blob/fix/font-style-issues/packages/app/src/components/App.css

PR #635, opened 2020-10-06
chore: try different format for plurals for transifex

jenniferarnesen:
I am manually setting the plural format in the pot file (pot file is normally auto-generated when source files are committed), to see if this results in transifex updating the strings. The current format is ignored by transifex.

I also removed the extract part from the build. Only translation json file generation makes sense at build time.

I don't want to trigger a release with this commit, only want to trigger transifex.  Note that if transifex recognizes this format, then translators will get notified and new PO files generated.

If this format works, then we'll have to address the extract process to see how we can generate the correct format.


Philip-Larsen-Donnelly (approved):
Looks OK to me.

I checked out the PR branch and tested pushing the file manually to transifex. It was successful, and the plurals are editable in transifex.

image

PR #657, opened 2020-11-30
fix: replace unsafe componentWillReceiveProps method

edoardo:
Required by https://github.com/dhis2/analytics/pull/655.

The new FileMenu uses `@dhis2/ui` components.
Conditional rendering is used to open a Popover component, since it doesn't have a `open` prop that can be toggled.
The problem was that the FavoritesDialog need to be added to the DOM with the `open` and `refreshData` props already set to `true`.
The old `componentWillReceiveProps`, never triggered, because those
props don't change in that case, they are just set when the `FavoritesDialog`
component renders.
`componentDidMount` is executed after the component is rendered, so the
condition for `open` and `refreshData` is checked also when using the
component without toggling the `open` prop.
`componentDidUpdate` is executed when a prop changes, this is triggered
when using the old approach of toggling the open prop.

### DIFF
--- a/packages/favorites-dialog/src/Favorites.js
+++ b/packages/favorites-dialog/src/Favorites.js
@@ -14,9 +14,24 @@ import EnhancedTable from './EnhancedTable';
 import { fetchData } from './actions';
 
 class Favorites extends Component {
-    componentWillReceiveProps(nextProps) {
+    componentDidMount() {
         // fetch data only the first time the dialog is open
-        if (nextProps.open && (!nextProps.dataIsLoaded || nextProps.refreshData)) {
+        if (
+            this.props.open &&
+            (!this.props.dataIsLoaded || this.props.refreshData)
+        ) {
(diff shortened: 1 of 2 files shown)


packages/favorites-dialog/src/Favorites.js
```diff
@@ -14,9 +14,24 @@ import EnhancedTable from './EnhancedTable';
 import { fetchData } from './actions';
 
 class Favorites extends Component {
-    componentWillReceiveProps(nextProps) {
+    componentDidMount() {
         // fetch data only the first time the dialog is open
```
jenniferarnesen, line 18: Seems like an unnecessary comment.

jenniferarnesen (approved):

packages/favorites-dialog/src/Favorites.js
edoardo: True, it does not apply anymore.
Fixed.

PR #681, opened 2021-02-19
chore: move ItemSelector component from analytics to PeriodSelector

jenniferarnesen:
This fix allows us to remove the dependency to @dhis2/analytics. And since PeriodSelector is the only component in dhis2 to still be using the ItemSelector we can remove the ItemSelector from analytics.

These are the changes to ItemSelector (from its latest version in analytics)
* convert styles from styled-jsx to css-in-js. Pseudoclass selectors were moved to PeriodSelector.css A few !importants were needed in order to avoid a larger refactor of the component styling
* removed data-test attributes since they aren't used anymore, and were causing warnings while running jest tests

https://user-images.githubusercontent.com/6113918/108528111-584c6680-72d3-11eb-9d5d-a018bad9d730.mov

### DIFF
--- a/examples/create-react-app/src/App.js
+++ b/examples/create-react-app/src/App.js
@@ -74,6 +74,9 @@ class App extends Component {
                     <HeaderBarExample d2={this.state.d2} />
                     <p>Look at the top of the screen...</p>
 
+                    <h2>Period selector</h2>
+                    <PeriodSelector d2={this.state.d2} />
+
                     <h2>Rich Text</h2>
                     <RichTextExample />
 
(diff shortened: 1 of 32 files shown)


martinkrulltott (approved):
Gave this a quick review as I don't see the need for an in-depth review of mostly copy/pasted code from Analytics, this component isn't used anywhere AFAIK and I simply don't have time 🤷‍♂️  but overall LGTM. If this old PeriodSelector is actually used somewhere we should test it properly within that app (or even consider refactor to use the new period selector in Analytics?).

jenniferarnesen:
> Gave this a quick review as I don't see the need for an in-depth review of mostly copy/pasted code from Analytics, this component isn't used anywhere AFAIK and I simply don't have time 🤷‍♂️ but overall LGTM. If this old PeriodSelector is actually used somewhere we should test it properly within that app (or even consider refactor to use the new period selector in Analytics?).

It is still in use in older versions of dhis2 (2.34, 2.33). I'll make sure to test those versions before they are released

PR #682, opened 2021-02-24
feat(expressionManager): add expression functions

Birkbjo:
Fixes https://jira.dhis2.org/browse/DHIS2-7581.

Note this goes into v29 of d2-ui, used by Maintenance-app.


varl:
@Birkbjo i think we ended up putting the entire expression manager into the maintenance app to avoid this, so closing it.

PR #688, opened 2021-04-19
fix: add theme provider to translation and sharing dialogs [DHIS2-10143]

jenniferarnesen:
Part of fix for <DHIS2-10143>.

Add option to use the mui theme provider directly in the Sharing and Translation dialogs so that the parent app does not need to do this. In some cases the app will no longer need to have material-ui as a direct dependency (i.e., dashboard-app). Related PR:  https://github.com/dhis2/dashboard-app/pull/1704/)

Although it doesn't seem to matter if the theme is provided at multiple points (i.e., both at the app level and the component level), I added an option to avoid any potential errors I might not have found.

To test, remove the mui theme from the cra App.js, and change the insertTheme property in sharing.js and translation.js

### DIFF
--- a/packages/sharing-dialog/src/SharingDialog.component.js
+++ b/packages/sharing-dialog/src/SharingDialog.component.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import { MuiThemeProvider, createMuiTheme } from '@material-ui/core/styles';
 import { mui3theme } from '@dhis2/d2-ui-core';
 import Dialog from '@material-ui/core/Dialog';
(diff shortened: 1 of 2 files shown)


packages/sharing-dialog/src/SharingDialog.component.js
```diff
@@ -224,8 +226,22 @@ class SharingDialog extends React.Component {
...
-            </div>
-        );
+            </Fragment>
+        )
+    }
+
+    render() {
+        if (this.props.insertTheme) {
+            return (
+                <div>
```
edoardo, line 236: Is this `<div>` necessary?

edoardo (approved):

packages/sharing-dialog/src/SharingDialog.component.js
jenniferarnesen: done

PR #692, opened 2021-04-21
fix: add and propagate insertTheme prop (DHIS2-10143)

edoardo:
Fixes narrow dialogs issue related to [DHIS2-10143](https://jira.dhis2.org/browse/DHIS2-10143) .

- fix the narrow dialog issue due to missing MUI theme

Other than TranslationsDialog and SharingDialog, other components need the same fix:
- FileMenu
- FavoritesDialog
- Interpretations

### DIFF
--- a/packages/favorites-dialog/src/FavoritesDialog.js
+++ b/packages/favorites-dialog/src/FavoritesDialog.js
@@ -1,6 +1,8 @@
 import React, { Component } from 'react';
 import { Provider } from 'react-redux';
 import PropTypes from 'prop-types';
+import { MuiThemeProvider, createMuiTheme } from '@material-ui/core/styles';
+import { mui3theme } from '@dhis2/d2-ui-core';
 
 import Favorites from './Favorites';
 import configureStore from './configureStore';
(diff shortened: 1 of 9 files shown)


packages/favorites-dialog/src/FavoritesDialog.js
```diff
@@ -51,14 +53,43 @@ class FavoritesDialog extends Component {
...
     }
+
+    render() {
+        if (this.props.insertTheme) {
+            return (
+                <MuiThemeProvider
+                    theme={createMuiTheme(
+                        // override the lg width to make it wider
+                        Object.assign({}, mui3theme, {
+                            overrides: {
```
jenniferarnesen, line 64: I'm concerned that doing it like this will delete all the other overrides (like DialogTitle, DialogContent, DialogActions, MenuItem, Select). One way to avoid this is by using immutability-helper: https://github.com/kolodny/immutability-helper (look at the `update` function)

jenniferarnesen (approved):

packages/favorites-dialog/src/FavoritesDialog.js
edoardo: Fixed by cloning the `mui3theme` object and overriding only the `paperWidthLg` key.

PR #693, opened 2021-05-20
feat: disable parts of interpretations and sharing components when offline [DHIS2-10874]

jenniferarnesen:
Part of fix for https://jira.dhis2.org/browse/DHIS2-10874

* Interpretations: When the containing app is offline, then the input areas of the Interpretation are covered with an overlay containing text "Not available offline". Also, the subscription bell (in the description area) is disabled and has a tooltip on hover.
* Sharing: Place an overlay over everything but the CLOSE button
 
I decided not to introduce dhis2/ui as a dependency and instead copied css from the dhis2/ui ComponentCover and Center components. The Mui Tooltip also has the dhis2/ui Tooltip styling.

Some html and css changes were needed to get the tooltip on the subscription bell to work correctly. (like switching from `float` to `flexbox`.

Before:
image

After:
Online: (notice that the subscription bell has more space around it and the description text does not wrap around it):
image

Offline:
The subscription bell:
image

The interpretation area with inputs and other interactions (some but not all needing connectivity) is completely covered.
image

Another example where there are a few interpretations:
image

Sharing Dialog offline:
image

### DIFF
--- a/packages/interpretations/src/components/DetailsPanel/Description.js
+++ b/packages/interpretations/src/components/DetailsPanel/Description.js
@@ -1,4 +1,4 @@
-import React, { Fragment } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import i18n from '@dhis2/d2-i18n';
 import { Parser as RichTextParser } from '@dhis2/d2-ui-rich-text';
--- a/packages/interpretations/src/components/DetailsPanel/Description.js
+++ b/packages/interpretations/src/components/DetailsPanel/Description.js
@@ -24,7 +24,7 @@ export const Description = ({ displayDescription, isToggled, onToggleDescription
         <p style={{fontStyle: 'italic'}}>{description}</p>;
 
     return (
-        <Fragment>
+        <div style={{display: 'flex', flexDirection: 'column'}}>
             {DescriptionElement}
             {displayDescription.length > descriptionMaxLength && (
                 <Link
@@ -32,7 +32,7 @@ export const Description = ({ displayDescription, isToggled, onToggleDescription
                     label={showMoreLessLabel}
                 />
             )}
-        </Fragment>
+        </div>
     );
 };
 
--- a/packages/interpretations/i18n/en.pot
+++ b/packages/interpretations/i18n/en.pot
@@ -5,8 +5,26 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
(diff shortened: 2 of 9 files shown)


packages/interpretations/src/components/DetailsPanel/Description.js
```diff
@@ -24,15 +24,15 @@ export const Description = ({ displayDescription, isToggled, onToggleDescription
         <p style={{fontStyle: 'italic'}}>{description}</p>;
 
     return (
-        <Fragment>
+        <div style={{display: 'flex', flexDirection: 'column'}}>
```
martinkrulltott, line 27: use a class instead of inline `style`?

martinkrulltott (commented):
Looks good as is, but curious to why `isOnline` was chosen instead of the opposite, `isOffline`. My thinking is 
`1)` that "no state" (i.e. `undefined`) should be the same as `false`, whereas here `undefined` falls back to `true`, which is a bit counter-intuitive IMHO. 
`2)` Also, since the state condition checks only apply to `!` states, (e.g. `if(!this.props.isOnline)`) but not the true cases, the makes me lean even more in favour of a `isOffline` prop instead (e.g `if(this.props.isOffline) //do offline stuff`). But maybe this is part of a bigger scope that I'm not seeing.
`3)` And I also think from an API perspective `isOffline` is more semantic than `isOnline` for the consumer / app.

packages/interpretations/src/components/DetailsPanel/Description.js
jenniferarnesen: I had decided to keep the surface area of changes low since this component is legacy and will be replaced eventually (don't know when though). Inline styles are already used in many places in the Interpretations components.

jenniferarnesen:
> Looks good as is, but curious to why `isOnline` was chosen instead of the opposite, `isOffline`. My thinking is
> `1)` that "no state" (i.e. `undefined`) should be the same as `false`, whereas here `undefined` falls back to `true`, which is a bit counter-intuitive IMHO.
> `2)` Also, since the state condition checks only apply to `!` states, (e.g. `if(!this.props.isOnline)`) but not the `true` cases, the makes me lean even more in favour of a `isOffline` prop instead (e.g `if(this.props.isOffline) //do offline stuff`). But maybe this is part of a bigger scope that I'm not seeing.
> `3)` And I also think from an API perspective `isOffline` is more semantic than `isOnline` for the consumer / app.

I thought about this as well. Ultimately I went with `isOnline` because that is what is used throughout the dashboard app, from the signature of the hook that app-platform will be providing (I've been told). But I agree with your point and maybe it doesn't matter that d2-ui components use different language.

martinkrulltott (approved):

cooper-joe (approved):

PR #695, opened 2021-05-21
fix: test add workflow for pushing to PRs

jenniferarnesen:
Run the install, build, lint and test workflows on pull requests. 

Fix test for Details component (and use Snapshot instead of manually checking for things)


varl:
Removing the `push` event means that the workflow will _not_ run when the commit is pushed to master (through the merge), and that means that the `jobs.publish` will not run on master, and nothing will be published.

Usually we use the `push` event instead of the `pull_request` event because a `push` event is triggered when a commit is pushed to any branch (including branches that are in a PR), and `pull_request` only triggered when a PR is changed.

My question is if those consequences are intentional, and that the intent is to trigger a `workflow_dispatch` event manually to trigger a release instead of releasing when merging to `master`?

jenniferarnesen:
> Removing the `push` event means that the workflow will _not_ run when the commit is pushed to master (through the merge), and that means that the `jobs.publish` will not run on master, and nothing will be published.
> 

> Usually we use the `push` event instead of the `pull_request` event because a `push` event is triggered when a commit is pushed to any branch (including branches that are in a PR), and `pull_request` only triggered when a PR is changed.
> 
> My question is if those consequences are intentional, and that the intent is to trigger a `workflow_dispatch` event manually to trigger a release instead of releasing when merging to `master`?

The current behaviour is actually that the developer has to manually trigger the workflow run to publish. So this PR adds new workflow runs on PRs. 

I started this PR by using `push`. The reason that I switched away from `push` is that the developer can merge a PR that might not have the npm version bumped. Bumping the npm version is a manual step that is easily forgotten (looking at myself here - e.g.: https://github.com/dhis2/d2-ui/pull/693).

varl:
@jenniferarnesen Ah! Right you are. Thanks for refreshing my memory. I remember now that I actually added the original with the `workflow_dispatch` workaround ... :facepalm:

varl (approved):

PR #724, opened 2022-04-04
fix: show new interpretation form when user/group has access DHIS2-12617

edoardo:
Fixes [DHIS2-12617](https://jira.dhis2.org/browse/DHIS2-12617)

- fix bug when checking if a user belongs to a group

This bug prevented to verify that the user belongs to a group and thus had access to write
a new interpretation.

Before:
Screenshot 2022-04-01 at 18 14 07

After:
Screenshot 2022-04-01 at 18 25 06


janhenrikoverland:
Update `users` variable name.

janhenrikoverland (approved):

PR #746, opened 2022-11-30
fix: add support for event visualizations

edoardo:

- add support for event visualizations
- fix URL for DV when using the View in app button


janhenrikoverland (approved):
LGTM

